PR #345, opened 2022-04-27
Project filter search

aniket-k-kaushik:
## Notion card
https://saeloun.notion.site/Filter-on-Projects-list-Page-7437280cf75d4a6ab33dd51221443fb9

Added Search for projects
Added Client filter for projects 
Added User filter for projects

Optimised project index page query, previously we were making 11 DB request, reduced it to 2

API : http://127.0.0.1:3000/internal_api/v1/projects?search=Project_1_Client_1_saeloun_India&user_id[]=1&client_id[]=1&billable=false

````json
{
    "projects": [
        {
            "id": 1,
            "name": "Project_1_Client_1_saeloun_India",
            "clientName": "client_1_saeloun_India",
            "isBillable": false,
            "minutesSpent": 10560.0
        }
    ],
    "clients": [
        {
            "id": 1,
            "name": "client_1_saeloun_India"
        },
        {
            "id": 2,
            "name": "client_2_saeloun_India"
        },
        {
            "id": 5,
            "name": "mMicrosoft"
        }
    ],
    "teamMembers": [
        {
            "id": 1,
            "userName": "Vipul A M"
        },
        {
            "id": 4,
            "userName": "Keshav Biswa"
        },
        {
            "id": 5,
            "userName": "Rohit Joshi"
        },
        {
            "id": 3,
            "userName": "Akhil G Krishnan"
        },
        {
            "id": 2,
            "userName": "Supriya Agarwal"
        }
    ]
}
````
## Type of change

### Checklist:

### CODE REVIEW

File: app/controllers/internal_api/v1/projects_controller.rb [hunk 1]
```diff
@@ -33,8 +33,17 @@ def destroy
 
   private
 
-    def projects
-      @_projects ||= current_company.projects.kept
+    # def projects
```
sudeeptarlekar, line 36: nit: remove if not being used anywhere

app/models/company.rb
```diff
@@ -35,6 +35,45 @@ class Company < ApplicationRecord
...
+    minutes_spent = timesheet_entries.group(:project_id).sum(:duration)
+    query = project_list.ransack({ name_or_client_name_or_is_billable_cont: search })
+    project_list = query.result
+    project_ids = project_list.ids.uniq
+    project_ids.map do |id|
+      billable_array = []
+      project_name_array = []
+      client_name_array = []
+      project_list.each do |project|
+        if id == project.id
```
sudeeptarlekar, line 49: I see that we are using only first element from these arrays, check if we can stop further processing if we find project.

Please add `TODO: Optimize` for this operation, will check later when we have a time

spec/models/company_spec.rb
```diff
@@ -157,5 +157,198 @@
...
+        create(:project_member, user_id: user_1.id, project_id: project_2.id)
+        create(:project_member, user_id: user_2.id, project_id: project_1.id)
+        create(:project_member, user_id: user_2.id, project_id: project_2.id)
+        create_list(:timesheet_entry, 5, user: user_1, project: project_1)
+        create_list(:timesheet_entry, 5, user: user_1, project: project_2)
+        create_list(:timesheet_entry, 5, user: user_2, project: project_1)
+        create_list(:timesheet_entry, 5, user: user_2, project: project_2)
+      end
+
+      def project_detail_list(client_id = nil, user_id = nil, billable = nil, search)
```
sudeeptarlekar, line 183: Why have we written same method again? Can't we access `company.project_list`

spec/models/company_spec.rb
```diff
@@ -157,5 +157,198 @@
...
+            clientName: client_name_array[0],
+            isBillable: billable_array[0],
+            minutesSpent: minutes_spent[id]
+          }
+        end
+      end
+
+      context "when no filters or search are applied" do
+        it "returns list of all projects" do
+          user_id, client_id, billable, search = nil
```
sudeeptarlekar, line 212: Please use `let` for these assignments, you can then tweak single value in context based on requirement.
e.g. you can only tweak `search` in next context. This way it will become much more cleaner.
Thoughts?

Please follow same comments for another method added

sudeeptarlekar (changes requested):

app/models/company.rb
aniket-k-kaushik: I, akhil and keshav have already tried it, we couldn't find another way to optimise it. I will added TODO for now

app/controllers/internal_api/v1/projects_controller.rb
aniket-k-kaushik: yes

spec/models/company_spec.rb
aniket-k-kaushik: yes, but in one of the examples, we have all the filters, which means we will have 11 lets, which rubocop doesn't allows, so keshav suggested this way.

app/models/company.rb
sudeeptarlekar: Log the ticket in Notion as well for same.

spec/models/company_spec.rb
aniket-k-kaushik: It's the model spec and we are testing project_list method only.

sudeeptarlekar: But why have we written duplicate method logic, if method is returning something, hardcode response and check if in spec

aniket-k-kaushik: we are using factory bot and faker gem,

app/controllers/internal_api/v1/projects_controller.rb
sudeeptarlekar: ?

aniket-k-kaushik: done

PR #351, opened 2022-04-28
Added Payments Setting tab in the settings 

aniket-k-kaushik:
## Notion card
### Checklist:


app/views/payments_setting/index.html.erb
```diff
@@ -0,0 +1 @@
+<H1>Code</H1>
```
keshavbiswa, line 1: I'm not sure what this is supposed to be?

app/controllers/payments_setting_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class PaymentsSettingController < ApplicationController
```
keshavbiswa, line 3: Shouldn't this be PaymentSettings controller?

config/routes.rb
```diff
@@ -62,6 +62,9 @@ def draw(routes_name)
   get "payments/*path", to: "payments#index", via: :all
   get "payments", to: "payments#index"
 
+  get "payments_setting/*path", to: "payments_setting#index", via: :all
+  get "payments_setting", to: "payments_setting#index"
```
keshavbiswa, line 66: I'm not understanding why are we doing this and not 
```ruby
resource :payment_settings, only: [:index]
```

Is there a specific use case I'm missing?

keshavbiswa (commented):
Added some comments.

app/controllers/payments_setting_controller.rb
aniket-k-kaushik: I can change it to paymentSettings,

PR #353, opened 2022-04-30
Fixes unit_amount for zero_decimal_currencies

mohd-anas-ansari:
## Preview

Earlier: 
Screenshot 2022-04-30 at 6 03 54 PM

After fix: 
Screenshot 2022-04-30 at 6 03 20 PM

## Type of change

### Checklist:

### DIFF
--- a/app/services/invoice_payment/checkout.rb
+++ b/app/services/invoice_payment/checkout.rb
@@ -45,7 +45,7 @@ module InvoicePayment
                   name: invoice.invoice_number,
                   description:
                 },
-                unit_amount: calculate_amount
+                unit_amount: invoice.unit_amount(company.base_currency)
               },
               quantity: 1
             }],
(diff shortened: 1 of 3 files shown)


apoorv1316:
@shivamsinghchahar Can you please review this?

rohitjoshixyz (approved):
The code looks good, add a service spec for both scenarios zero decimal and normal currency

app/services/invoice_payment/checkout.rb
```diff
@@ -55,5 +55,14 @@ def checkout!
             cancel_url:
           })
       end
+
+      def calculate_amount
```
rohitjoshixyz, line 59: We can keep the method name as `unit_amount`

app/services/invoice_payment/checkout.rb
```diff
@@ -55,5 +55,14 @@ def checkout!
...
+        zero_decimal_currencies = %w[BIF CLP DJF GNF JPY KMF KRW MGA PYG RWF UGX VND VUV XAF XOF XPF]
+        if zero_decimal_currencies.include?(currency)
+          amount = invoice.amount.to_i
+        else
+          amount = (invoice.amount * 100).to_i
+        end
```
shivamsinghchahar, line 65: We are using `money` gem, could you explore if there's a better way to handle it with that? If there is then we should have a method on the invoice model and use that, since this will be used at a lot of places, not just this particular service.

shivamsinghchahar (changes requested):

shivamsinghchahar (approved):
LGTM🚀

PR #383, opened 2022-05-17
Employee dropdown

abinashpa:
## Preview
https://www.loom.com/share/64f3149c1e1a4526a07a89c5faa22259

## Type of change

      not work as expected)

### Checklist:


app/controllers/internal_api/v1/timesheet_entry_controller.rb
```diff
@@ -8,7 +8,7 @@ class InternalApi::V1::TimesheetEntryController < InternalApi::V1::ApplicationCo
 
   def index
     timesheet_entries = policy_scope(TimesheetEntry)
-    timesheet_entries = timesheet_entries.where(user_id: params[:user_id] || current_user.id).during(
+    timesheet_entries = timesheet_entries.where(user_id: params[:uid] || current_user.id).during(
```
rohitjoshixyz, line 11: Why not `user_id`?

app/controllers/time_tracking_controller.rb
```diff
@@ -5,7 +5,10 @@ class TimeTrackingController < ApplicationController
   skip_after_action :verify_authorized
 
   def index
-    is_admin = current_user.is_admin?
+    is_admin = current_user.has_role? [:owner, :admin], current_company
```
rohitjoshixyz, line 8: Prefer using policy method

rohitjoshixyz, line 8: Use policy

app/controllers/time_tracking_controller.rb
```diff
@@ -5,7 +5,10 @@ class TimeTrackingController < ApplicationController
...
+    user_id = current_user.id
+    employees = is_admin ? current_user.current_workspace.users.select(:id, :first_name, :last_name) : [current_user]
```
rohitjoshixyz, line 10: We can use `current_company.user.select(...)` directly

rohitjoshixyz, line 10: We can use `current_company.user.select(...)` directly

app/controllers/time_tracking_controller.rb
```diff
@@ -5,7 +5,10 @@ class TimeTrackingController < ApplicationController
...
+
     clients = current_company.clients.includes(:projects)
     projects = {}
     clients.map { |client| projects[client.name] = client.projects }
```
rohitjoshixyz, line 14: We could use
`projects = clients.map{|client| [client.name, client.projects]}.to_h`

Question: What if we have the same client name for two clients? The value will get overwritten

rohitjoshixyz, line 14: We could use
`projects = clients.map{|client| [client.name, client.projects]}.to_h`

Question: What if we have the same client name for two clients? The value will get overwritten

app/javascript/src/apis/timesheet-entry.ts
```diff
@@ -4,7 +4,7 @@ const path = "/timesheet_entry";
 
 const create = async params => axios.post(path, params);
 
-const list = async (from, to) => axios.get(`${path}?from=${from}&to=${to}`);
+const list = async (from, to, uid) => axios.get(`${path}?from=${from}&to=${to}&uid=${uid}`);
```
rohitjoshixyz, line 7: Can you verify what happens if an employee(non-admin) manually enters the UID through the URL? We should be good since policy_scopw will filter all the timesheet_entries

rohitjoshixyz, line 7: Can you verify what happens if an employee(non-admin) manually enters the UID through the URL? We should be good since policy_scopw will filter all the timesheet_entries

rohitjoshixyz (commented):
Few suggestions

supriya3105:
@abinashpa Please check review comments from Rohit

app/controllers/time_tracking_controller.rb
abinashpa: Duplicate.

app/controllers/time_tracking_controller.rb
abinashpa: Yes, it will be overwritten.

abinashpa: Yes, it will be overwritten.

abinashpa: Duplicate comment.

app/javascript/src/apis/timesheet-entry.ts
abinashpa: Response will be like this:
`{ entries: {} }`

abinashpa: Duplicate comment.

app/controllers/time_tracking_controller.rb
abinashpa: Duplicate comment.

app/controllers/time_tracking_controller.rb
abinashpa: Yes, we can.

abinashpa:
@apoorv1316 Added the fixes for the issue (not fetching other employees' entries). Can you verify it?

apoorv1316:
@abinashpa Can you please look into this bug? https://www.loom.com/share/2245f4789a5e44ac80eef6e3998fd0e9

rohitjoshixyz (approved):
LGTM

apoorv1316 (approved):

PR #415, opened 2022-05-26
[BugFix]Create customer for connected account instead of platform

apoorv-mishra:
## Notion card
NA
## Summary

Fix to create customer for connected account instead of the platform. Refer [docs](https://stripe.com/docs/connect/authentication) for the right ways to make API calls for connected accounts.

## Type of change

      not work as expected)

### Checklist:

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -121,6 +121,8 @@ class Client < ApplicationRecord
           metadata: {
             platform_id: id
           }
+        }, {
+          stripe_account: stripe_connected_account.account_id
         })
 
       update!(stripe_id: customer.id)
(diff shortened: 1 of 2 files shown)


app/models/client.rb
```diff
@@ -121,6 +121,8 @@ def register_on_stripe!
           metadata: {
             platform_id: id
           }
+        }, {
+          stripe_account: stripe_connected_account.account_id
```
akhilgkrishnan, line 125: can you add a test for this?

apoorv1316 (approved):
Tested and payment was successful in local

shivamsinghchahar (approved):
@apoorv-mishra Nice catch! LGTM

app/models/client.rb
apoorv-mishra: Added.

PR #454, opened 2022-06-07
Team member Delete Api

harshaanNihal:
## Notion card
https://www.notion.so/saeloun/API-development-for-delete-team-member-ba3b8f43bff5483a8c3bb456bc28a4cc

## Summary
- Add internal api for Team page with destroy Method
- Add specs for Api team_controller#destroy method

## Type of change

### Checklist:


app/controllers/internal_api/v1/team_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class InternalApi::V1::TeamController < InternalApi::V1::ApplicationController
```
rohitjoshixyz, line 3: NIt: `TeamController` -> `TeamsController`

app/controllers/internal_api/v1/team_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  def destroy
+    authorize current_user, policy_class: TeamPolicy
```
rohitjoshixyz, line 5: We do not need to specify the policy class explicitly if the controller and policy have the same name

akhilgkrishnan, line 5:
```suggestion
    authorize Team
```

app/controllers/internal_api/v1/team_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    user.discard!
```
akhilgkrishnan, line 6: what is the response when a team member does not discard?

app/controllers/internal_api/v1/team_controller.rb
akhilgkrishnan, line 3:
```suggestion
class InternalApi::V1::TeamsController < InternalApi::V1::ApplicationController
```
Can we use a plural name

spec/requests/internal_api/v1/teams/destroy_spec.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
```
akhilgkrishnan, line 1: Add a test for negative condition, when user is not discarded properly

akhilgkrishnan (commented):
@harshaanNihal Added some comments

spec/requests/internal_api/v1/teams/destroy_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+require "rails_helper"
+
+RSpec.describe "InternalApi::V1::Team#destroy", type: :request do
+  let(:company) { create(:company) }
+  let(:user) { create(:user, current_workspace_id: company.id) }
```
rohitjoshixyz, line 7: You can rename this `user` as `admin` or `admin_user` and in the "when user is unauthorised" call it employee or employee_user to be more specific

rohitjoshixyz (changes requested):
I have a few concerns with this PR and hence stopped reviewing further. We are not checking if the user being deleted is part of other companies or not.
Discarding the user directly would discard the `company_users` associated with other companies making the user unable to login into other workspaces
This scenario is not tested in the specs as well.

What is the exact requirement of this API? We can probably split this API into two.
- User destroy API which will completely remove the access to Miru (For all the workspaces)
- Company User destroy API which will destroy one particular `company_user` (For a particular workspace)

Also when we say team member it is not immediately clear if we are talking about the company_user or the larger user record cc @supriya3105 @harshaanNihal @akhilgkrishnan

akhilgkrishnan:
> I have a few concerns with this PR and hence stopped reviewing further. We are not checking if the user being deleted is part of other companies or not. Discarding the user directly would discard the `company_users` associated with other companies making the user unable to login into other workspaces This scenario is not tested in the specs as well.
> 
> What is the exact requirement of this API? We can probably split this API into two.
> 
> * User destroy API which will completely remove the access to Miru (For all the workspaces)
> * Company User destroy API which will destroy one particular `company_user` (For a particular workspace)
> 
> Also when we say team member it is not immediately clear if we are talking about the company_user or the larger user record cc @supriya3105 @harshaanNihal @akhilgkrishnan

Yes I agree with you.

harshaanNihal:
> I have a few concerns with this PR and hence stopped reviewing further. We are not checking if the user being deleted is part of other companies or not. Discarding the user directly would discard the company_users associated with other companies making the user unable to login into other workspaces This scenario is not tested in the specs as well.
> 
> What is the exact requirement of this API? We can probably split this API into two.
> 
> User destroy API which will completely remove the access to Miru (For all the workspaces)
> Company User destroy API which will destroy one particular company_user (For a particular workspace)
> Also when we say team member it is not immediately clear if we are talking about the company_user or the larger user record cc @supriya3105 @harshaanNihal @akhilgkrishnan

Agree we should discard the `conmpany_user` instead of `user` in this scenario.
Initially I was trying to port prewritten [team_controller](https://github.com/saeloun/miru-web/blob/develop/app/controllers/team_controller.rb) to internal API which was discarding user directly hence I overlooked this scenario.

Moving the PR to Draft for now

app/controllers/internal_api/v1/team_controller.rb
harshaanNihal: fixed

app/controllers/internal_api/v1/team_controller.rb
harshaanNihal: in case `user.discard!` is not successful it will raise an Exception, which will get rescued [here](https://github.com/saeloun/miru-web/blob/2fc6b3043e49d0b88b5a9eaed5e47b1e3d34b863/app/controllers/concerns/error_handler.rb#L9)

spec/requests/internal_api/v1/teams/destroy_spec.rb
harshaanNihal: Added a Model spec for negative condition and mocking the failure negative condition in controller specs

harshaanNihal:
I have kept the controller name singular `team` instead of `teams`.
The controller will mainly manage the `company_users` and it's not tied to a specific model  i.e. `Team`.

Since each company will technically have only one team (company users belong to that company). As Team controller is not associated with any particular model / resource, but rather handle a set of functionality.  In this case, the pluralization doesn't make any sense to me.

Please let me know if you think otherwise and I will update the controller name to a plural form.
@rohitjoshixyz @akhilgkrishnan

rohitjoshixyz:
>Since each company will technically have only one team (company users belong to that company). As Team controller is not associated with any particular model / resource, but rather handle a set of functionality. In this case, the pluralization doesn't make any sense to me.

That makes sense, considering this and our previous discussion since this controller is only going to handle the `company_user` model's destroy action let's name it as `company_users_controller`. What do you think?

harshaanNihal:
There is already a company_users_controller present. [for now it is been used for project Page ]
The Team_controller will be handling functionality specific to team page. [It is already implemented for ERB files]. We are porting it to APIs so that the team page can be ported into react components.

beside destroy method controller will have index update and probably edit.

Maybe once after the Team page is ported we can merge both `team` and `company_user` controller into single controller. if they share the functionality.

rohitjoshixyz (approved):
@harshaanNihal 
Summarizing our discussion here.

The original rails controller's (`app/controllers/team_controller`) implementation itself is wrong. 
Instead of discarding the `User` it should ideally discard only the current_workspace's `CompanyUser` record.
According to the latest changes I asked you to make here, this PR will do that correctly.

But as you said the intent of this PR is to port the functionality of the rails controller into an internal API to be used for React we are in a fix. 😓

I won't block this PR since this is an internal API but we need to address and add API for both:
1. Discarding the User record
2. Discarding the Company User record
And name the controllers accordingly on priority as a follow-up

Please check and update the language and grammar used in the spec description. It should be consistent.

Overall nice work with the specs. LGTM 👍

akhilgkrishnan (approved):

PR #457, opened 2022-06-09
Time entry report export

sudeeptarlekar:
## Notion card
https://www.notion.so/saeloun/Time-entry-Report-Export-as-CSV-7c53b4d80f9942588129ffe705e6aa85

## Summary
Ability to export time entry report as CSV of PDF

## Preview
https://user-images.githubusercontent.com/23317691/172738060-cd93d281-5786-4105-8c1a-929aeaddcce9.mp4

## Type of change

### Checklist:


supriya3105 (approved):

app/javascript/src/components/Reports/index.tsx
```diff
@@ -84,6 +85,20 @@ const Reports = () => {
...
+    reports.download(type, `?${queryParams}`)
+      .then(response => window.URL.createObjectURL(new Blob([response.data])))
+      .then(url => {
+        const link = document.createElement("a");
+        link.href = url;
+        link.setAttribute("download", `report.${type}`);
+        document.body.appendChild(link);
+        link.click();
+      })
+      .catch(error => console.error(error));
```
apoorv-mishra, line 99: We should replace this with `async/await` like we're doing at other places. Helps maintain consistency.

sudeeptarlekar: Fixed

supriya3105:
@apoorv1316 Please test this on priority.

sudeeptarlekar:
Where you tested this?

supriya3105:
@apoorv1316 Was this discussed with Sudeep after today's stand up call ?

app/services/report/generate_csv.rb
```diff
@@ -0,0 +1,29 @@
...
+module Report
+  class GenerateCsv
+    attr_reader :entries
+
+    def initialize(entries)
+      @entries = entries
+    end
+
+    def process
+      CSV.generate(headers: true) do |csv|
```
akhilgkrishnan, line 14: @sudeeptarlekar Can we move the report generation to a worker? Otherwise, this entire CSV generation works synchronously, And it will cause heavy load in the server.

sudeeptarlekar: We can, but how are we thinking of making it available for download to user?

sudeeptarlekar: Imo, synchronous download for now will work as there won't be much data, though I agree this should be asynchronous, but we can fix this in future, it will be little efforts when we decide how we will let user download a file.

Thoughts?

akhilgkrishnan: I'm okay with that.

app/controllers/internal_api/v1/reports_controller.rb
```diff
@@ -8,6 +8,17 @@ def index
...
   end
 
+  def download
+    authorize :report
+
+    entries = reports.map { |e| e[:entries] }.flatten
+
+    respond_to do |format|
+      format.csv { send_data Report::GenerateCsv.new(entries).process }
+      format.pdf { send_data Report::GeneratePdf.new(reports).process }
```
akhilgkrishnan, line 18:
```suggestion
      format.csv { send_data Report::GenerateCsv.new(entries).process, filename:  "createddate_companyname_miru_report"}
      format.pdf { send_data Report::GeneratePdf.new(reports).process, filename: "createddate_companyname_miru_report" }
```

apoorv1316 (approved):

app/controllers/internal_api/v1/reports_controller.rb
sudeeptarlekar: We are adding filename from React.

supriya3105 (approved):

PR #461, opened 2022-06-09
Creates EmploymentDetail model

mohinid:
## Notion card

## Summary
1. Creates employment_details table with appropriate columns.
2. Creates 1-1 association between CompanyUser and EmploymentDetail model.
3. Adds validation for the EmploymentDetail model.
4. Add RSpec for both EmploymentDetail and CompanyUser models.

## Type of change

## How Has This Been Tested?
Added validations & RSpecs for User and EmploymentDetail models.

### Checklist:


spec/models/employment_detail_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+RSpec.describe EmploymentDetail, type: :model do
+  subject(:employment_detail) { build(:employment_detail) }
+
+  describe "Validations" do
+    describe "Associations" do
+      it { is_expected.to belong_to(:user) }
+    end
+
+    describe "validate presence of" do
```
rohitjoshixyz, line 13:
```suggestion
    describe "Presence" do
```

spec/models/employment_detail_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      it { is_expected.to validate_presence_of(:employment_type) }
+      it { is_expected.to validate_presence_of(:joined_at) }
+      it { is_expected.to validate_presence_of(:employee_id) }
+    end
+
+    describe "validate comparisons" do
+      it "resignation date should be later than joining date" do
+        expect(employment_detail.resigned_at).to be > employment_detail.joined_at
+      end
+    end
```
rohitjoshixyz, line 24: What happens when either of them is `nil`? Add a spec for each

spec/factories/employment_details.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :employment_detail do
+    user
+    designation { "SDE" }
+    employment_type { "Salaried" }
+    joined_at { Faker::Date.between(from: "2020-01-01", to: "2021-01-01") }
+    resigned_at { Date.today }
```
rohitjoshixyz, line 9: This factory will always create a resigned employee by default.
Let's keep `resigned_at` nil by default

rohitjoshixyz (commented):
@mohinid Added a few suggestions

spec/models/employment_detail_spec.rb
mohinid: Joined_at cannot be nill. I have validation added for the same. Will add test for nil resigned_at date.

app/models/user.rb
```diff
@@ -60,6 +60,7 @@ class User < ApplicationRecord
   has_many :identities, dependent: :delete_all
   has_one_attached :avatar
   rolify strict: true
+  has_one :employment_detail, dependent: :destroy
```
akhilgkrishnan, line 63: can you move this on line:62

akhilgkrishnan (commented):
@mohinid I have got a couple of doubts related to this flow because, As per the Miru workflow the User can be part of multiple companies and the employee details vary with the position handled in each company. So we should add a scope for the company model.

app/models/user.rb
mohinid: done

mohinid:
> @mohinid I have got a couple of doubts related to this flow because, As per the Miru workflow the User can be part of multiple companies and the employee details vary with the position handled in each company. This EmploymentDetails model hasn't given an idea about the company which the employee is associated with. So we should add a scope for the company model.

Hi @akhilgkrishnan , let's discuss today.

rohitjoshixyz:
@akhilgkrishnan Good catch: we could associate the `employment_details` table with the `company_users` table instead of the `user` model, which should address this concern @mohinid

mohinid:
Removed User reference and added reference of company_detail with employment_detail. @akhilgkrishnan  @rohitjoshixyz 
All tests passed. Please review.

Screenshot 2022-06-13 at 3 14 13 PM

rohitjoshixyz (approved):
LGTM 🎉

akhilgkrishnan (approved):
LGTM. Thanks @mohinid for making the requested changes

PR #466, opened 2022-06-13
Buy plan page

Shruti-Apte:
## Notion card

https://www.notion.so/saeloun/Develop-the-Plan-details-page-5dae88a09c1f48b0a9d22707161ac81a

## Summary

- Designed Buy plan page 
- Functions for calculating amount added

## Preview

https://www.loom.com/share/43341e2b03b64265a6a469053847759d

## Type of change

### Checklist:


app/javascript/src/components/PlanDetails/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+          <div className="modal__form flex-col">
+            <div className="field mt-4 flex justify-between">
+              <div className="field_with_errors py-2">
+                <label className="font-normal text-xs leading-4 text-miru-dark-purple-1000">
+                  Number of Team Members
+                </label>
+              </div>
+              <input
+                type="text"
```
onkar-saeloun, line 48: Why we are taking input type as text and not as number. If we use input type number you don't need to write logic for parsing the text as well you can set min property too.

app/javascript/src/components/PlanDetails/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+const PlanDetails = () => {
+  const [teamMembersCount, setTeamMembersCount] = useState<number>(1);
+  const [planTotal, setPlanTotal] = useState<number>(15);
+
+  const handleCountChange = (e) => {
+    if (e.target.value) {
+      const count = parseInt(e.target.value);
+      if (isNaN(count)) {
+        Toastr.error("Please enter valid data");
```
onkar-saeloun, line 14: Ideally toaster should be displayed when you need to show some notifications or after an API call to show response status. We shouldn't use toaster on onChange method here you can handle error state like 
`const [error, setError] = useState<string>("")` 
and show it in `p` tag with danger colour below your inputbox to show error.

app/javascript/src/components/PlanDetails/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+        setTeamMembersCount(1);
+        setPlanTotal(15);
+      } else if (count == 0) {
+        Toastr.error("Number of team members can not be zero");
```
onkar-saeloun, line 18: You can use Yup validation here or simply use min property if feasible.

app/javascript/src/components/PlanDetails/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+        setTeamMembersCount(1);
+        setPlanTotal(15);
+      } else {
+        setTeamMembersCount(count);
+        setPlanTotal(count * 5 + 10);
```
onkar-saeloun, line 23: Please don't hardcode the amount lets say your basic_plan_charge = 10 and team_member_charge = 5 then make a file in constant folder and keep it over there. Your initial state value should be also come from the calculation 
`const [planTotal, setPlanTotal] = useState<number>(basic_plan_charge + team_member_charge);`
If the values will come from database then please ignore the comment.

onkar-saeloun (changes requested):
@Shruti-Apte Please let me know in case of any issue.

app/javascript/src/components/PlanDetails/index.tsx
Shruti-Apte: Yes.. But if I use type=number down and up arrows are also getting displayed which are not there in our designs. 
Is there any other way to remove those default css other than webkits.

app/javascript/src/components/PlanDetails/index.tsx
Shruti-Apte: Okay

app/javascript/src/components/PlanDetails/index.tsx
onkar-saeloun: Ohh then you can simply use Yup validations its simple to use you can add validations over there also.

Shruti-Apte:
Becasue of multiple files changed (rebasing issue) creating a new PR.

PR #473, opened 2022-06-14
Adds new personal detail fields to Users table

mohinid:
## Notion card

## Summary
1. Added new columns(personal_email_id, date_of_birth & social_accounts) to Users table.
2. DATA MIGRATION- Updated default empty string values to Social Account URLs(github_url, linkedin_url) for existing Data.
3. Added "UserDefaultSocialAccountVerifier" - verification file for data migration.
4. Added rspec for new changes in User model. 

## Preview
Screenshot 2022-06-14 at 4 19 45 PM

## Type of change

## How Has This Been Tested?
Tested on Local Rails console.

### Checklist:


app/models/user.rb
```diff
@@ -26,11 +27,13 @@
 #  reset_password_sent_at :datetime
 #  reset_password_token   :string
 #  sign_in_count          :integer          default(0), not null
+#  social_accounts        :jsonb
```
apoorv-mishra, line 30: I would suggest to have a separate model for `social_accounts` with a `has_many :social_accounts` association on `User` model. Fits well with the relational model.

mohinid: @rohitjoshixyz  can you please comment.

app/models/user.rb
```diff
@@ -62,7 +65,11 @@ class User < ApplicationRecord
   has_one_attached :avatar
   rolify strict: true
 
+  # Social account details
+  store_accessor :social_accounts, :github_url, :linkedin_url
+
   # Validations
+  before_validation :set_default_social_accounts, on: :create
```
rohitjoshixyz, line 72: Can you comment on why did you choose the `before_validation` callback? I think we can have it `after_create`

Also, we are not checking if the URLs already exist in the object. Wouldn't this overwrite the URLs we actually want to save?

app/models/user.rb
apoorv-mishra: Cool 👍

db/data/20220614094240_add_personal_detail_fields_to_users.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative("./verifiers/user_default_social_account_verifier.rb")
```
rohitjoshixyz, line 3: Do we need this to be in a separate directory? We could just write the method in the migration itself. Could you post any blog or tutorial you followed for this over here for my reference?

app/models/user.rb
```diff
@@ -111,6 +118,13 @@ def send_devise_notification(notification, *args)
     super
   end
 
+  def set_default_social_accounts
+    self.social_accounts = {
+      "github_url": "",
+      "linkedin_url": ""
+    }
+  end
+
```
rohitjoshixyz, line 127: @mohinid Creating empty strings is fine. `before_validation :set_default_social_accounts, on: :create` can be replaced with `before_create` as it would be equivalent and I think it should be `after_create` cc @akhilgkrishnan

app/models/user.rb
akhilgkrishnan: I think, In this case, I don't think we need an after_create. The problem with after_create it actually write the DB multiple times. 
One for creating the user entry, another write for updating the social_account. before_validation or before_create is suitable for here, because they are just initializing the value and not creating an entry for the DB.

app/models/user.rb
akhilgkrishnan: Yes, before_create also works

app/models/user.rb
```diff
@@ -26,11 +27,13 @@
 #  reset_password_sent_at :datetime
 #  reset_password_token   :string
 #  sign_in_count          :integer          default(0), not null
+#  social_accounts        :jsonb
 #  unconfirmed_email      :string
 #  created_at             :datetime         not null
 #  updated_at             :datetime         not null
 #  current_workspace_id   :bigint
 #  invited_by_id          :bigint
+#  personal_email_id      :string
```
akhilgkrishnan, line 36: @mohinid Still got confused with this field. What's the exact difference of the actual email and this one.

app/models/user.rb
rohitjoshixyz: Let's go ahead with `before_create` 👍

app/models/user.rb
mohinid: This is the personal Email Id as of the user just for reference. Like an alternative email.

app/models/user.rb
mohinid: @akhilgkrishnan  @rohitjoshixyz what is the final conclusion? before_create or before_validation?

akhilgkrishnan: Go with before_create

mohinid: @akhilgkrishnan @rohitjoshixyz I am using **"after_initialize"** with a check of new_record. It creates social_accounts with empty URLs because 'before_create' is generating an empty social_accounts object.

Tested this by updating the github_url of the first record, as well as running RSpecs. Here is the SS.

Screenshot 2022-06-14 at 8 55 05 PM

rohitjoshixyz (approved):
LGTM, I like the `before_initialize` callback 👍

PR #481, opened 2022-06-15
Creates previous_employment_details table

mohinid:
## Notion card

## Summary
1. Created previous_employment_details table with fields: company_name and role.
2. Added associations to EmploymentDetails and CompanyUsers model.
3. Added. validations on fields.
4. Added RSpec
5. Tested manually on Console.

## Type of change

## How Has This Been Tested?
Manually verified associations and validations on the Rails console.

### Checklist:

### Preview of ERD
Employee detail schema ERD


db/schema.rb
```diff
@@ -135,7 +138,9 @@
     t.bigint "client_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "external_view_key"
     t.index ["client_id"], name: "index_invoices_on_client_id"
+    t.index ["external_view_key"], name: "index_invoices_on_external_view_key", unique: true
```
rohitjoshixyz, line 143: This looks like an unrelated schema change

rohitjoshixyz (commented):
Check schema file, rest LGTM

db/schema.rb
mohinid: Yes it took some invoice changes!

apoorv-mishra (changes requested):
Could you resolve conflicts?

akhilgkrishnan (commented):
@mohinid I've some concerns about this association. The think the previous company details should be directly associated with the User model.  For a user, the previous employment should same for each company and it does not vary from company to company.
Eg: Consider, Currently a user belongs to 2 companies A and B. Same user has worked for 3 companies before. As per your flow, the user has 2 employment_details ( for company A and Company B) and each employee details have 3 same previous employees details.

mohinid:
> @akhilgkrishnan Well, it makes sense that we associate the model to users since we are storing only `name` and `role` for now. That won't change for any particular individual. It will also reduce a join so let's go for it 👍 @mohinid

Thanks, @rohitjoshixyz, @akhilgkrishnan I will change it.

keshavbiswa:
> , Currently a user belongs to 2 companies A and B. Same user has worked for 3 companies before

Agreed. Shouldn't the same be true for employee_details table?

The user would have the same employee_details regardless of whichever company he is part of.

mohinid:
> > , Currently a user belongs to 2 companies A and B. Same user has worked for 3 companies before
> 
> Agreed. Shouldn't the same be true for employee_details table?
> 
> The user would have the same employee_details regardless of whichever company he is part of.

Hi @keshavbiswa Not really. Please check the fields of employment_details table on ERD. It has employee_id, designation etc which is specific to each company.

keshavbiswa:
> 

So, if current_user has two `company_users` records, this would mean the current_user would have two seperate `employment_details`?

keshavbiswa:
Okay, understood. But if that's the case then, is there a need to create employment_details table? I mean wouldn't it be easier to add these attributes to company_users table directly?

mohinid:
> Okay, understood. But if that's the case then, is there a need to create employment_details table? I mean wouldn't it be easier to add these attributes to company_users table directly?

I think from what I understand a good DB has normalization and having separate table for similar data is a better approach.

mohinid:
Updated reference from CompanyUser to User as discussed. 
Also, tested on local.
Will update the ERD in next PR!

spec/models/user_spec.rb
```diff
@@ -16,6 +16,7 @@
     it { is_expected.to have_many(:identities).dependent(:delete_all) }
     it { is_expected.to have_many(:project_members).dependent(:destroy) }
     it { is_expected.to have_many(:timesheet_entries) }
+    it { is_expected.to have_many(:previous_employment_details) }
```
keshavbiswa, line 19:
```suggestion
    it { is_expected.to have_many(:previous_employment_details).dependent(:destroy }
```

app/models/employment_detail.rb
```diff
@@ -25,6 +25,7 @@
 
 class EmploymentDetail < ApplicationRecord
   belongs_to :company_user
+  has_many :previous_employment_details
```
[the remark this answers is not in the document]
keshavbiswa: Also could you add a one-liner test for this association, just like the user_spec?

keshavbiswa (commented):
Added few suggestions for missing tests, rest LGTM!

app/models/user.rb
```diff
@@ -62,6 +62,7 @@ class User < ApplicationRecord
   has_many :timesheet_entries
   has_many :identities, dependent: :delete_all
   has_one :wise_account, dependent: :destroy
+  has_many :previous_employment_details, dependent: :destroy
```
keshavbiswa, line 65: Could you also add a spec for this one?

keshavbiswa (approved):
Thanks for making all the changes. Looks good to me!

mohinid:
> Thanks for making all the changes. Looks good to me!

Thanks a lot @keshavbiswa for your time!! :)

rohitjoshixyz (approved):
LGTM

PR #516, opened 2022-06-23
Refactor route to consider report type

shalapatil:
## Summary
For reports page, we will have multiple types of reports. So, now api routes will look like "......./reports/time_entry" instead of "....../reports"
Did refactoring for other related code also

## Type of change

      not work as expected)

### Checklist:


app/services/time_entry_report/filters.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module Report
+module TimeEntryReport
```
rohitjoshixyz, line 3: Just a suggestion and you can take the call if you want to change it or not. I would prefer to keep the service namespaces consistent with the controller namespacing in which they are being used.
i.e: `app/controllers/reports/time_entry_controller` could have its services defined in `app/services/reports/time_entry/result_service.rb` and so on

spec/requests/internal_api/v1/clients/show_spec.rb
```diff
@@ -43,7 +43,7 @@
       send_request :get, internal_api_v1_client_path(client_1)
     end
 
-    it "is not permitted to view time entry report" do
+    it "is not permitted to view cliet details" do
```
rohitjoshixyz, line 46: Typo
```suggestion
    it "is not permitted to view client details" do
```

app/controllers/internal_api/v1/reports/time_entry_controller.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class InternalApi::V1::ReportsController < InternalApi::V1::ApplicationController
+class InternalApi::V1::Reports::TimeEntryController < InternalApi::V1::ApplicationController
```
rohitjoshixyz, line 3: As per rails conventions, we should name the controller in plural form. `TimeEntriesController`

rohitjoshixyz (commented):
A couple of suggestions

rohitjoshixyz (approved):
Thank you for working on the refactoring. LGTM 🎉

PR #522, opened 2022-06-24
Prevent employee from invitating team members and fixed invitation flow

gowsik-ragunath:
## Notion card

https://www.notion.so/saeloun/Update-Deviceinvitetable-3a653cf772104331bd7ddee7d91e46b8

## Summary

- Added a pundit policy to prevent employees from inviting team members
- Updated the invitation controller create action to invite both new and existing users.

## Preview

https://www.loom.com/share/07c179ada15543fe8708fef5c38415a3

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
      not work as expected)

### Checklist:

- [X] I have manually tested all workflows
- [X] I have performed a self-review of my own code
- [X] I have added automated tests for my code


app/models/company_user.rb
```diff
@@ -36,6 +36,6 @@ class CompanyUser < ApplicationRecord
   belongs_to :user
 
   # Validations
-  validates :designation, :employment_type, :joined_at, :employee_id, presence: true
-  validates :resigned_at, comparison: { greater_than: :joined_at }, unless: -> { resigned_at.nil? }
+  # validates :designation, :employment_type, :joined_at, :employee_id, presence: true
+  # validates :resigned_at, comparison: { greater_than: :joined_at }, unless: -> { resigned_at.nil? }
```
akhilgkrishnan, line 40: Why do we need this change.

app/views/devise/mailer/invitation_instructions.html.erb
```diff
@@ -8,12 +8,11 @@
...
 </ul>
 
-<p><%= t("devise.mailer.invitation_instructions.confirm") %></p>
-
 <% if @token.present? %>
+  <p><%= t("devise.mailer.invitation_instructions.confirm") %></p>
   <p><%= link_to t("devise.mailer.invitation_instructions.confirm_link"), accept_invitation_url(@resource, invitation_token: @token) %></p>
 <% else %>
-  <p><%= link_to t("devise.mailer.invitation_instructions.confirm_link"), root_url %></p>
+  <p><%= link_to "Visit workspace", root_url %></p>
```
akhilgkrishnan, line 15: but for the first-time user they have to set the password right?. In that case I feel it should be the `Accept Invitation`.

app/models/company_user.rb
gowsik-ragunath: In the backend, we have this validation but the frontend is not updated to get all these data, until the frontend is updated we should disable all these validations.

[Slack message ref.](https://saelounus.slack.com/archives/C033ZCDGJ10/p1656066640643929)

app/views/devise/mailer/invitation_instructions.html.erb
gowsik-ragunath: Yeah, for non-existing users the `if` block will be executed and the link in the mail will have `Yes, it's me, confirm my email address`, when the user clicks on that link it will take them to the set password page.

For existing users the `else` block will be executed and the link in the email will have `Visit workspace`, when clicked it will take them to the root path.

supriya3105:
@keshavbiswa @rohitjoshixyz Please review this PR as discussed in yesterday's stand up call

PR #533, opened 2022-06-29
Invitation migration - part 1

gowsik-ragunath:
## Notion card

https://www.notion.so/saeloun/Refactor-the-Invite-user-tables-4d3fb482b71b462586be8fb0bb5436da

## Summary

- Added migration for the invitation.
- Wrote a rake for moving devise invitation to invitation table.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
      not work as expected)

### Checklist:

- [X] I have performed a self-review of my own code
- [X] I have added automated tests for my code


lib/tasks/one_off_invitation_migration.rake
```diff
@@ -0,0 +1,23 @@
+desc 'Move all the pending devise invitable invitation to invitation table'
+task move_devise_invitation_to_invitation_table: :environment do
+  User.where(invitation_accepted_at: nil).where.not(invitation_token: nil).find_each do |user|
+    user.roles.each do |role|
+      expiration_date = if user.invitation_sent_at.is_a? ActiveSupport::TimeWithZone
+        user.invitation_sent_at + 14.days
+      else
+        Time.now + 14.days
```
keshavbiswa, line 8:
```suggestion
        Time.current + 14.days
```

lib/tasks/one_off_invitation_migration.rake
```diff
@@ -0,0 +1,23 @@
...
+        recipient_email: user.email,
+        role: role.name.to_sym,
+        token: user.invitation_token,
+        expired_at: expiration_date,
+        first_name: user.first_name,
+        last_name: user.last_name
+      )
+    end
+  end
+end
```
keshavbiswa, line 23: Shouldn't this be a data migration? and not a one-time migration task?

spec/factories/invitations.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :invitation do
+    company
+    user
+  end
```
keshavbiswa, line 7: We could also add other attributes for the factory.

app/models/invitation.rb
```diff
@@ -0,0 +1,35 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (company_id => companies.id)
+#  fk_rails_...  (sender_id => users.id)
+#
+class Invitation < ApplicationRecord
+  enum role: [:owner, :admin, :employee, :book_keeper]
+
+  belongs_to :company
+  belongs_to :sender, foreign_key: "sender_id", class_name: "User"
```
keshavbiswa, line 34: You could also add validations for presence of token, role and recipient email.

spec/models/invitation_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Invitation, type: :model do
+  describe "Associations" do
```
keshavbiswa, line 6: You could also add specs for validations for presence of token, role and recipient email.

app/models/invitation.rb
keshavbiswa: Why `sender_id` and not `user_id` ?

app/models/timesheet_entry.rb
```diff
@@ -7,7 +7,7 @@
 #  id          :bigint           not null, primary key
 #  bill_status :integer          not null
 #  duration    :float            not null
-#  note        :text             default("")
+#  note        :text
```
keshavbiswa, line 10: This is not part of this PR. I don't know from where this keeps popping up 🤦🏻

keshavbiswa (commented):
Added few comments and suggestions.

lib/tasks/one_off_invitation_migration.rake
keshavbiswa: You can make use of the [data_migrate](https://github.com/ilyakatz/data-migrate) gem since we're already using that. If the data migration is created before the db migration then we can run `db:migrate:with_data` to run them sequentially.

gowsik-ragunath: Cool - didn't know that we are using data_migrate gem, will move that logic.

app/models/timesheet_entry.rb
gowsik-ragunath: Yeah I'm not sure, but [this PR](https://github.com/saeloun/miru-web/commit/b193d2de8cdf47ca3041c181d26defb0d5fa7c38#diff-19998e355dcb83d3ecf9dc070e20b8bb3cb00a56cebd13b949c10fdfb75dbd32) has the migration to change the note column as nullable and set a default value for it.

Even in this PR timesheet_entry model has
```#  note        :text```
And not
```note        :text             default("")```

It might be an issue in the `annotate` gem or because of us using a different Postgres version.

[Here](https://github.com/ctran/annotate_models/issues/278) is a link to a closed issue that is similar to the issue that we are facing.

Comment (keshavbiswa) [reply app/models/timesheet_entry.rb:10]: Let's merge this change in your pr and end this nightmare once and for all.

File: app/models/invitation.rb (hunk 4 above)
Comment (gowsik-ragunath) [reply app/models/invitation.rb]: Actually, the foreign key is not needed here because of the migration so, I removed it.

[ref:](https://railsguides.net/foreign-key-for-custom-table-name/)

File: spec/models/invitation_spec.rb (hunk 5 above)
Comment (gowsik-ragunath) [reply spec/models/invitation_spec.rb:8]: Added tests for validations.

File: app/models/invitation.rb [hunk 7]
```diff
@@ -30,6 +30,11 @@
...
   enum role: [:owner, :admin, :employee, :book_keeper]
 
+  # Associations
   belongs_to :company
-  belongs_to :sender, foreign_key: "sender_id", class_name: "User"
+  belongs_to :sender, class_name: "User"
+
+  # Validations
+  validates :recipient_email, :role, :token, presence: true
+  validates_uniqueness_of :token
```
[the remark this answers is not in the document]
gowsik-ragunath: Yeah, do we need to add this in this PR? Because this PR is only for adding the migration, we will be adding this callback in later PR in which we will implement the whole invitation flow.

keshavbiswa (approved):
LGTM!

akhilgkrishnan (approved):

PR #600, opened 2022-08-02
Delete autogenerated file and code

apoorv1316:
## Summary
- Deleted index.js which was created while installing the old versions of cypress. From cypress 10 onwards we don't need the 'plugins/index.js' file. So removed it along with a piece of code from 'cypress.config.ts'.

## Type of change

      not work as expected)

### Checklist:


cypress.config.ts
```diff
@@ -3,11 +3,6 @@ import { defineConfig } from 'cypress'
...
   projectId: 'upc13x',
   e2e: {
-    // We've imported your old cypress plugins here.
-    // You may want to clean this up later by importing these.
-    setupNodeEvents(on, config) {
-      return require('./cypress/plugins/index.js')(on, config)
-    },
     baseUrl: 'http://localhost:3000',
     specPattern: 'cypress/e2e/**/*.{js,jsx,ts,tsx}',
     execTimeout:1800000,
```
mohinid, line 8: why is this exec timeout so high?

apoorv1316: I just copied it from the last config 😬. Anyway, we can remove this because we are not using cy.exec() command anywhere in the app.

apoorv1316: @mohinid I have removed the execTimeout like we discussed. Please approve the PR.

mohinid (approved):

PR #609, opened 2022-08-09
Combine previous employments APIs for Employee and Admin

mohinid:
## Notion card
https://www.notion.so/saeloun/6afc9500b2cc42af86be55f37894fb09?v=b4528f26f4424af7beaf036262796cb3&p=0e05f8024ea94d3a99c62ac6a18e2656&pm=s

## Summary
Previously we had created 2 separate APIs for employee(with index and create to access own record) and admin (index, create for a specific user). Have a workaround to combine both cases in 1 API.

## Type of change

- [x] Enhancement

## How Has This Been Tested?
RSpec

### Checklist:


spec/requests/internal_api/v1/users/previous_employments/create_spec.rb
```diff
@@ -131,6 +131,25 @@
       end
 
       it "is successful" do
+        expect(response).to have_http_status(:ok)
```
keshavbiswa, line 134: Since we have covered all the cases in our policy files we may only use simple cases for our request specs.

keshavbiswa (approved):

spec/requests/internal_api/v1/users/previous_employments/create_spec.rb
mohinid: hey no some of them are not covered like in valid Ids and unsuccessful responses. hence covered all.

PR #610, opened 2022-08-09
Update Addresses API

mohinid:
## Notion card
https://www.notion.so/saeloun/6afc9500b2cc42af86be55f37894fb09?v=b4528f26f4424af7beaf036262796cb3&p=e9441e11d56649c8af3d91b01a6c70b0&pm=s

## Summary
Changes:
- adds respective routes for company and users using concerns
- adds index and create actions
- fixes show and update actions
- updates policy file to test both addressable types
- adds/updates RSpecs

## Preview
SS attached for routes and postman

## Type of change

- [x] Enhancement
- [x] Breaking change (fix or feature that would cause existing functionality to
      not work as expected)

## How Has This Been Tested?

Via Postman & Rspecs on local

### Checklist:


mohinid:
@keshavbiswa I will work on updating the specs tomorrow. You can meanwhile have a look at the new implementation and policy tests.

app/controllers/concerns/addressable_concern.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module AddressableConcern
+  extend ActiveSupport::Concern
+end
```
keshavbiswa, line 5: Why have we created an empty addressable concern?

app/controllers/internal_api/v1/addresses_controller.rb
```diff
@@ -1,6 +1,26 @@
 # frozen_string_literal: true
 
 class InternalApi::V1::AddressesController < InternalApi::V1::ApplicationController
+  include AddressableConcern
```
keshavbiswa, line 4: And why are we including empty concern here? What does it achieve?

app/controllers/internal_api/v1/addresses_controller.rb
```diff
@@ -1,6 +1,26 @@
...
+
+  before_action :set_addressable_type, :set_addressable
```
keshavbiswa, line 6: We don't need to set addressable_type here, we can get addressable using params[:addressable_type].

app/controllers/internal_api/v1/addresses_controller.rb
```diff
@@ -1,6 +1,26 @@
...
+    addresses = @addressable.addresses
+    render :index, locals: { addresses: }, status: :ok
+ end
+
+  def create
+    authorize @addressable, policy_class: AddressPolicy
+    address = @addressable.addresses.new(
+      address_params.merge(
+        addressable_type: @addressable_type,
+        addressable_id: @addressable.id))
```
keshavbiswa, line 19: This is incorrect, you can have a look and try and figure out why.

app/controllers/internal_api/v1/addresses_controller.rb
```diff
@@ -14,8 +34,20 @@ def update
...
+      if @addressable_type == "User"
+        @addressable = User.find(params[:user_id])
+      else
+        @addressable = Company.find(params[:company_id])
+      end
+    end
+
     def address
-      @address ||= Address.find(params[:id])
+      @address ||= @addressable.addresses.find(params[:id])
```
keshavbiswa, line 50: Whenever we're finding a record with ID, it doesn't matter if we do Address.find() or addressable.addresses.find() because ultimately we're returning the record with ID. So this should be 

```ruby
@address ||= Address.find(params[:id])
```

app/policies/address_policy.rb
```diff
@@ -1,21 +1,37 @@
 # frozen_string_literal: true
 
 class AddressPolicy < ApplicationPolicy
+  def index?
+    if record.class.name == "User"
```
keshavbiswa, line 5:
```suggestion
    if record.is_a?(User)
```

app/policies/address_policy.rb
```diff
@@ -1,21 +1,37 @@
...
+      (record.employed_at?(user.current_workspace_id) &&
+      user_is_admin_or_owner_in_current_workspace?) || record == user
+    else
```
keshavbiswa, line 8:
```suggestion
    elsif record.is_a?(Company)
```

config/routes/internal_api.rb
```diff
@@ -41,8 +41,13 @@
     resources :employments, only: [:index]
     resources :timezones, only: [:index]
 
+    concern :AddressableConcern do |options|
+      resources :addresses, options
+    end
```
keshavbiswa, line 46: This is incorrect, look up concerns in Routes.

Also now I understand, why you may have created ControllerConcern. That is not needed at all.

app/controllers/internal_api/v1/addresses_controller.rb
mohinid: Hi I dont think this is incorrect. here the @addressable is the user or company that we have passed in the URL. and we need to store that ID in addresses row.

app/controllers/internal_api/v1/addresses_controller.rb
mohinid: resolved

app/controllers/concerns/addressable_concern.rb
mohinid: removed

app/controllers/internal_api/v1/addresses_controller.rb
mohinid: fixed. thanks!!!

config/routes/internal_api.rb
```diff
@@ -41,8 +41,13 @@
     resources :employments, only: [:index]
     resources :timezones, only: [:index]
 
+    concern :addressable do |options|
+      resources :addresses, options
+    end
```
keshavbiswa, line 46:
```suggestion
    concern :addressable do
      resources :addresses, only: %i[index create show update]
    end
```

config/routes/internal_api.rb
```diff
@@ -41,8 +41,13 @@
...
+
     resources :companies, only: [:index, :create, :update] do
```
keshavbiswa, line 48:
```suggestion
    resources :companies, only: [:index, :create, :update], concerns: :addressable do
```

config/routes/internal_api.rb
```diff
@@ -41,8 +41,13 @@
...
       resource :purge_logo, only: [:destroy], controller: "companies/purge_logo"
+      concerns :addressable, addressable_type: "Company"
```
keshavbiswa, line 50: I don't know what this line would do.

config/routes/internal_api.rb
```diff
@@ -82,10 +87,9 @@
     resources :users do
```
keshavbiswa, line 87:
```suggestion
    resources :users, concerns: :addressable do
```

config/routes/internal_api.rb
```diff
@@ -82,10 +87,9 @@
...
       resources :previous_employments, only: [:create, :index, :show, :update], controller: "users/previous_employments"
       resources :devices, only: [:create, :index, :show, :update], controller: "users/devices"
+      concerns :addressable, addressable_type: "User"
```
keshavbiswa, line 90: Remove this line.

config/routes/internal_api.rb
mohinid: We have to set addressable_type according to the resource, hence using this. If we remove this line - we will not receive addressable_type in params.

keshavbiswa: > We have to set addressable_type according to the resource for checking the resource passed in index and create action, hence using this. If we remove this line - we will not receive addressable_type in params.

You don't need addressable_type, You can do something like this as well:
```ruby
if params[:user_id]
  @addressable = User.find(params[:user_id])
elsif params[:company_id]
  @addressable = Company.find(params[:company_id])
end
```

keshavbiswa (approved):

PR #612, opened 2022-08-10
Project page redirect bug

abinashpa:
## Notion card
[Notion Card](https://www.notion.so/saeloun/After-deleting-a-project-from-the-client-s-details-page-the-user-is-directed-to-the-projects-list-p-dd0902c97385426c9e297941ec8ae541)

## Summary
- After deleting a project from the client page it should redirect to the selected client page instead of /project.

## Preview
https://www.loom.com/share/7365d730eed0490487bb035a56086ad3

## Type of change

      not work as expected)

### Checklist:


supriya3105:
@abinashpa Can you add the link to the related notion ticket as well as add a loom video of the fix.

apoorv1316:
@abinashpa In the notion, it's mentioned that after deleting a project it should be redirected to the client's details page but here in the PR description, it's mentioned that after deleting the project it should be redirected to the client's page. So what is the actually expected behavior here? And please update the description on the notion or GitHub wherever it is required.

abinashpa:
@apoorv1316 Fixes have been added for redirect back to the same client page.

apoorv1316 (approved):
@abinashpa LGTM!

app/javascript/src/components/Projects/Modals/DeleteProject.tsx
```diff
@@ -12,7 +12,11 @@ const DeleteProject = ({ project, setShowDeleteDialog }: IProps) => {
       .then(() => {
         setTimeout(() => {
           setShowDeleteDialog(false);
-          window.location.assign(window.location.origin+"/projects");
+          if (window.location.pathname.includes("/clients")) {
```
ajinkyaa, line 15: why using `window.location`? 
Using routes there right?

app/javascript/src/components/Projects/Modals/DeleteProject.tsx
```diff
@@ -12,7 +12,11 @@ const DeleteProject = ({ project, setShowDeleteDialog }: IProps) => {
...
+            window.location.href = window.location.pathname;
+          } else {
+            window.location.assign(window.location.origin+"/projects");
+          }
```
akhilgkrishnan, line 19: Why do we need a redirection?

app/javascript/src/components/Projects/Modals/DeleteProject.tsx
```diff
@@ -12,7 +12,11 @@ const DeleteProject = ({ project, setShowDeleteDialog }: IProps) => {
       .then(() => {
         setTimeout(() => {
           setShowDeleteDialog(false);
```
akhilgkrishnan, line 14: If we are redirecting the page. Then there is no point of having this method.

akhilgkrishnan (approved):
`If we are using window.location that actually redirect the page and sent all requests to the backend. This is against the SPA`

I'm approving this, but we should consider for a refactor ASAP.

supriya3105:
@abinashpa we should be using the same method that we have implemented for clients page. 

@ajinkyaa Can you confirm which method should be used here.

ajinkyaa:
@akhilgkrishnan I had the same question when I was reviewing this last night.
it's needed because Abinash is redirecting it from the client page and routes are not set up on the project level. 

@supriya3105 if teams page is converted in react, can we have a call regarding setting up the routes on the project level?
that way we don't have to reload the application every time.

ajinkyaa (approved):

ajinkyaa (changes requested):
Please wait I am discussing react-router with Onkar.

ajinkyaa:
@abinashpa had a discussion with Harshaan. 
you will need to use react-router methods instead of window methods. 
We already have a route setup for both client and project.

abinashpa:
@ajinkyaa On /project and /clients/:id page uses different data formats so it's impossible to set data from DELETE /projects/:id so I have implemented it in the previous way you suggested.

abinashpa:
@apoorv1316 Please test /client and /project page.

supriya3105:
@ajinkyaa Can you review this PR is its implemented as per you discussion with Abinash.

ajinkyaa:
Closing this PR, @abinashpa will be raising PR with new branch.

PR #644, opened 2022-09-08
Added vendor folder to .gitignore

prasanthchaduvula:
## Notion card
https://www.notion.so/6afc9500b2cc42af86be55f37894fb09?v=b4528f26f4424af7beaf036262796cb3&p=c78c618d68cc450290fd7980d39695f3&pm=s

## Summary

Added /vendor/* folder to .gitignore as it creates a lot of untracked files when we do bundle install —path vendor/bundle.
Even for assets, it's better to use the javascript folder where by default assets will be loaded to the application.js file. With the vendor, if we want to add assets we have to manually require them in the application.js file. Please refer https://stackoverflow.com/a/24169365
Updated the gitignore cache which added test folders to gitignore. As we are using rspec.

## Type of change
- Added /vendor/* and !/vendor/.keep to .gitignore
- Ran git rm -r --cached .
- Updated the files that were meant to be added to .gitignore file
- .gitingore cache and ignored all the mentioned files in .gitignore file.

### Checklist:


prasanthchaduvula:
@keshavbiswa Please test it locally. As checks, CI checks are passing here but in local I am facing the below issue. Maybe my setup is wrong but still please test it before you merge it. In the above summary, I added my thoughts on the vendor folder as well.

```
Companies#new when user is an employee renders Company#new page
      Failure/Error: <%= javascript_pack_tag "org_timezones", "data-turbo-track": "reload", defer: true %>
      
      ActionView::Template::Error:
        Webpacker can't find org_timezones.js in /Users/prasanthchaduvula/Documents/saeloun/miru-web/public/packs-test/manifest.json. Possible causes:
        1. You want to set webpacker.yml value of compile to true for your environment
           unless you are using the `webpack -w` or the webpack-dev-server.
        2. webpack has not yet re-run to reflect updates.
        3. You have misconfigured Webpacker's config/webpacker.yml file.
        4. Your webpack configuration is not creating a manifest.
        Your manifest contains:
        {
        }
  ```

File: .gitignore [hunk 1]
```diff
@@ -41,6 +41,11 @@
 yarn-debug.log*
 .yarn-integrity
 
+# Rails specific ignores
+/vendor/*
```
keshavbiswa, line 45: Like I mentioned I would like the vendor folder to exists. Since this can be used to store vendor assets.

keshavbiswa (changes requested):

.gitignore
prasanthchaduvula: @keshavbiswa Vendor folder along with the .keep file exists here with the above changes. Please check the file changes where vendor folder files didn't change even after running `git rm -r --cached .`
```
vendor
  javascript
    .keep
  .keep
```

keshavbiswa: Yes but future additions in vendor folder will remain ignored which we don't want. For this PR we only want to ignore vendor/bundle part.

prasanthchaduvula: @keshavbiswa got it but please give it a thought for future sustainable development https://stackoverflow.com/a/24169365
I will make the requested changes

keshavbiswa: Yeah, I'm not saying we're going to add our assets to vendor folder, but there is a reason why Rails include a vendor folder, if we're going to ignore it in our gitignore file entirely, then it just won't make sense.

keshavbiswa (approved):
LGTM!

PR #648, opened 2022-09-09
Updated bin setup for sustainable web development

prasanthchaduvula:
## Notion card
https://www.notion.so/6afc9500b2cc42af86be55f37894fb09?v=b4528f26f4424af7beaf036262796cb3&p=77f0d72f009849f8a5229dbcde2a4c1b&pm=s

## Summary

Basically, bin/setup helps developers to set up the app easily without running the commands manually. With this PR there is no need to run the below commands manually. After running `bin/setup`, we can run `foreman start -f Procfile.dev` and can view the app on `localhost:3000`. To run tests we can use `bundle exec rspec`

```
Install gems
bundle install

Install node packages
yarn install

Setup ENV's
cp .env.example .env

Update `DATABASE_URL` in `.env` as per local `psql` creds. For example, if
    the user is `root` and password is `password`, change the variable as
    `DATABASE_URL="postgres://root:password@localhost/miru_web?encoding=utf8&pool=5&timeout=5000"`

Update `APP_BASE_URL` in `.env` to `localhost:3000`

Run `bin/rails db:create RAILS_ENV=development` to create the database
Run `bin/rails db:migrate RAILS_ENV=development` for migrations
Run `bin/rails db:seed` for populating the database with initial data

Tests
Run bin/rails db:create RAILS_ENV=test
Run bin/rails db:migrate RAILS_ENV=test
```

## Preview

https://www.loom.com/share/3075c380c5a44a7a96b337e945b34c07

## Type of change
- Added bin/yarn
- Updated .env.example and database.yml to support setting up a database without database URL in dev, test.
- Added rake setup to run tasks related to database
- Updated bin/setup to run bundle install, yarn install, create .env file, drop db, create db, run migrations, and seed the data.

### Checklist:

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -22,13 +22,8 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Executing yarn =="
   system!("bin/yarn")
 
-  # puts "\n== Copying sample files =="
-  # unless File.exist?("config/database.yml")
-  #   FileUtils.cp "config/database.yml.sample", "config/database.yml"
-  # end
-
-  # puts "\n== Preparing database =="
-  # system! "bin/rails db:prepare"
+  puts "\n== Setup ENV's =="
+  system! "cp .env.example .env"
 
   puts "\n== Executing rake setup =="
   system! "bundle exec rake setup"
--- a/bin/setup
+++ b/bin/setup
@@ -25,8 +25,12 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Setup ENV's =="
   system! "cp .env.example .env"
 
-  puts "\n== Executing rake setup =="
-  system! "bundle exec rake setup"
+  # Drop database to have a unique seed data otherwise db:seed will fail with validation errors.
+  puts "\n== Drop database =="
+  system! "bin/rails db:drop"
+
+  puts "\n== Preparing database =="
+  system! "bin/rails db:prepare"
 
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
--- a/bin/setup
+++ b/bin/setup
@@ -14,16 +14,24 @@ FileUtils.chdir APP_ROOT do
   # Add necessary setup steps to this file.
 
   puts "== Installing dependencies =="
-  system! "gem install bundler --conservative"
+  cmd = "gem install bundler --conservative"
+  puts "Executing #{cmd}"
+  system! cmd
   system("bundle check") || system!("bundle install")
+
+  puts "\n== Executing yarn =="
+  system!("bin/yarn")
 
   # puts "\n== Copying sample files =="
   # unless File.exist?("config/database.yml")
   #   FileUtils.cp "config/database.yml.sample", "config/database.yml"
   # end
 
-  puts "\n== Preparing database =="
-  system! "bin/rails db:prepare"
+  # puts "\n== Preparing database =="
+  # system! "bin/rails db:prepare"
+
+  puts "\n== Executing rake setup =="
+  system! "bundle exec rake setup"
 
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,18 +1,17 @@
+<% branch_name = `git symbolic-ref HEAD 2>/dev/null`.chomp.sub('refs/heads/', '') %>
+<% repository_name = `git rev-parse --show-toplevel`.split('/').last.strip %>
+
 default: &default
   adapter: postgresql
   encoding: unicode
-  # For details on connection pooling, see Rails configuration guide
-  # https://guides.rubyonrails.org/configuring.html#database-pooling
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  host: localhost
+  pool: 5
+  password:
 
 development:
   <<: *default
-  url: <%= ENV['DATABASE_URL'].gsub('?', '_development?') %>
+  database: <%= "#{repository_name}_development" %>
 
 test:
   <<: *default
-  url: <%= ENV['DATABASE_URL'].gsub('?', '_test?') %>
-
-production:
-  <<: *default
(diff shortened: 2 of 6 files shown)


bin/setup
```diff
@@ -14,16 +14,19 @@ FileUtils.chdir APP_ROOT do
...
+  puts "\n== Executing yarn =="
+  system!("bin/yarn")
 
-  puts "\n== Preparing database =="
-  system! "bin/rails db:prepare"
+  puts "\n== Setup ENV's =="
+  system! "cp .env.example .env"
+
+  puts "\n== Executing rake setup =="
+  system! "bundle exec rake setup"
```
keshavbiswa, line 29: The only difference I see from `rake setup` and `rails db:prepare` is that we're dropping the db everytime we're running bin/setup which I don't think should be the default here. If I want to reset my db, I can always drop my db before running bin/setup.

config/database.yml
```diff
@@ -1,17 +1,19 @@
+<% repository_name = Rails.application.class.module_parent_name.underscore %>
```
keshavbiswa, line 1: This assumes that our db's name is prefixed with the application name, but what if it's not?

bin/setup
```diff
@@ -14,16 +14,19 @@ FileUtils.chdir APP_ROOT do
...
-  system! "gem install bundler --conservative"
+  cmd = "gem install bundler --conservative"
+  puts "Executing #{cmd}"
+  system! cmd
   system("bundle check") || system!("bundle install")
 
-  # puts "\n== Copying sample files =="
-  # unless File.exist?("config/database.yml")
-  #   FileUtils.cp "config/database.yml.sample", "config/database.yml"
-  # end
```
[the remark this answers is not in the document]
prasanthchaduvula: @keshavbiswa I didn't understand the above comment. Please can you provide me more context

bin/setup
prasanthchaduvula: @keshavbiswa But every time when ever we run `bin/setup` we have to run `db:drop` otherwise it will through the active record uniqueness validation errors. example email should be unique. So if anyone wants to run `bin/setup `they have to run `db:drop` before doing it. So the best way will be to automate it instead of manually doing it.

```

== Command ["bin/rails db:prepare"] failed ==   We can run db:seed once gain with the current codebase as there unique validations and for that once gain developer has to follow all the commands and drop database  bin/rails db:seed   
Company Created
rails aborted!
ActiveRecord::RecordInvalid: Validation failed: Email Email ID already exists
/Users/prasanthchaduvula/Documents/saeloun/miru-web/db/seeds/02_users.rb:30:in `block in <main>'
/Users/prasanthchaduvula/Documents/saeloun/miru-web/db/seeds/02_users.rb:30:in `map'
/Users/prasanthchaduvula/Documents/saeloun/miru-web/db/seeds/02_users.rb:30:in `<main>'
/Users/prasanthchaduvula/Documents/saeloun/miru-web/db/seeds.rb:4:in `block in <main>'
/Users/prasanthchaduvula/Documents/saeloun/miru-web/db/seeds.rb:3:in `each'
/Users/prasanthchaduvula/Documents/saeloun/miru-web/db/seeds.rb:3:in `<main>'
Tasks: TOP => db:seed
(See full trace by running task with --trace)

```

config/database.yml
prasanthchaduvula: @keshavbiswa I verified and it's working fine. It creates a db named `miru_web_development `and earlier with the previous code our intention is to use the repo app name itself. If someone wants to override it they can customize the databse.yml but for regular developers, it won't create any problem as repo/app name `miru_web_development ` is the best practice and it also removes the hassle of updating the developer's username and password of `DATABASE_URL` in env file every time whenever they clone the repo. It's not a good idea to keep `DATABASE_URL` as it's intended to use in the deployment platforms like Heroku. The best way is to keep it out of the env file.
https://devcenter.heroku.com/articles/rails-database-connection-behavior

config/database.yml
```diff
@@ -1,17 +1,19 @@
...
-  # For details on connection pooling, see Rails configuration guide
-  # https://guides.rubyonrails.org/configuring.html#database-pooling
+  host: localhost
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  password:
 
 development:
   <<: *default
-  url: <%= ENV['DATABASE_URL'].gsub('?', '_development?') %>
+  database: <%= "#{repository_name}_development" %>
```
keshavbiswa, line 12:
```suggestion
  database: "miru_web_development"
```

config/database.yml
```diff
@@ -1,17 +1,19 @@
...
 
 test:
   <<: *default
-  url: <%= ENV['DATABASE_URL'].gsub('?', '_test?') %>
+  database: <%= "#{repository_name}_test" %>
```
keshavbiswa, line 16:
```suggestion
  database: "miru_web_test"
```

config/database.yml
keshavbiswa: Let's remove this.

bin/setup
keshavbiswa: Does everytime running bin/setup requires us to drop our db and seed the database again?

prasanthchaduvula: @keshavbiswa Yes because every time we are seeding data and without dropping the database our seed data is not unique and validation fails like email etc... That's the reason we have to drop the database every time for unique seed data.

keshavbiswa (approved):
LGTM!

db/seeds.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+return if Rails.env.production?
```
akhilgkrishnan, line 3: can you remove this for now? Because right now we don't have a separate environment for staging (Rails environment). So we are using the production.rb configuration for staging (For maintaining the replica of actual production). We can add this later by adding staging.rb configuration.

akhilgkrishnan (commented):
@prasanthchaduvula Added one comment. Rest LGTM

akhilgkrishnan (approved):
LGTM

PR #649, opened 2022-09-09
make date format dd.mm.yy consistent

apoorv1316:
## Notion card
https://www.notion.so/saeloun/6afc9500b2cc42af86be55f37894fb09?v=b4528f26f4424af7beaf036262796cb3&p=db86bf5e667b4734985d25928f157900&pm=s

## Summary
Make the date format 'DD.MM.YY' consistent throughout the app.
## Preview
Screenshot 2022-09-09 at 7 27 06 PM
Screenshot 2022-09-09 at 7 27 23 PM
Screenshot 2022-09-09 at 7 27 53 PM
Screenshot 2022-09-09 at 7 28 21 PM
Screenshot 2022-09-09 at 7 29 10 PM

## Type of change

      not work as expected)

### Checklist:

### DIFF
--- a/app/javascript/src/helpers/formattedDate.ts
+++ b/app/javascript/src/helpers/formattedDate.ts
@@ -1,0 +1,3 @@
+import dayjs from "dayjs";
+
+export const formattedDate = (date) => dayjs(date).format("DD.MM.YYYY");
--- a/app/javascript/src/components/Invoices/Edit/NewLineItemTable.tsx
+++ b/app/javascript/src/components/Invoices/Edit/NewLineItemTable.tsx
@@ -1,7 +1,8 @@
 import React from "react";
 
-import { getMaxIdx } from "../common/utils";
+import { formattedDate } from "helpers/formattedDate";
 
+import { getMaxIdx } from "../common/utils";
 const NewLineItemTable = ({
   lineItems, setLineItems,
   selectedLineItems, setSelectedLineItems,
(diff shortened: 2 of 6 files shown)


apoorv1316:
@supriya3105 I have made the date format in DD.MM.YY as per the requirement.

apoorv1316:
@prasanthchaduvula Please review it. @supriya3105 Let me know If I have missed any page where the date format is something other than the "DD.MM.YY"

prasanthchaduvula (approved):
@apoorv1316 Please check the date format working fine wherever it is not using the helper method. If the logic written in those components is not fine then use this helper method

app/javascript/src/helpers/formattedDate.ts
```diff
@@ -0,0 +1,3 @@
+import dayjs from "dayjs";
+
+export const formattedDate = (date) => dayjs(date).format("DD.MM.YYYY");
```
akhilgkrishnan, line 3: @apoorv1316 I think we are collecting the date format from the company. Can we use the company date format here? 
cc: @supriya3105

apoorv1316: @supriya3105 I think Akhil is right here. Because we are asking for the date format when we create a new company we should show the same date format throughout the app instead of making the date format DD.MM.YY. And if the user changes the date format it should be changed automatically at all places. Thoughts on this?

supriya3105: > @supriya3105 I think Akhil is right here. Because we are asking for the date format when we create a new company we should show the same date format throughout the app instead of making the date format DD.MM.YY. And if the user changes the date format it should be changed automatically at all places. Thoughts on this?

@apoorv1316 I agree. We should take the date format from the organization settings and apply it across the platform.

PR #652, opened 2022-09-12
Added error routes to the application

sudeeptarlekar:
Signed-off-by: sudeeptarlekar <sudeeptarlekar@gmail.com>

## Notion card
https://www.notion.so/saeloun/After-removing-a-team-member-from-an-organization-the-user-is-stil-able-to-login-54138da1e64540298474ca5eb2b4e04d

## Summary
Removed role from user against company when removed from the team. Missing designs, added simple error page for now.

## Preview
https://user-images.githubusercontent.com/23317691/189617250-287a1132-eb6e-4f19-bda1-5743120ebe89.mp4

## Type of change
- [ x ] Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [ x ] I have manually tested all workflows


akhilgkrishnan:
@supriya3105, @sudeeptarlekar I don't think this is a right behaviour, if a user has been deleted then we shouldn't allow then to sign in. We should block them in sign-in page itself.
But there are some scenarios we should take care of,
1. User belongs to multiple company, and he has removed from one company. Then he should able to login with the other company.
2. If a user belongs to only one company then we can disable the login.

LMK the thoughts
cc: @apoorv1316

akhilgkrishnan (commented):
@supriya3105, @sudeeptarlekar I don't think this is a right behaviour, if a user has been deleted then we shouldn't allow then to sign in. We should block them in sign-in page itself.
But there are some scenarios we should take care of,
1. User belongs to multiple company, and he has removed from one company. Then he should able to login with the other company.
2. If a user belongs to only one company then we can disable the login.

LMK the thoughts
cc: @apoorv1316

supriya3105:
> @supriya3105, @sudeeptarlekar I don't think this is a right behaviour, if a user has been deleted then we shouldn't allow then to sign in. We should block them in sign-in page itself. But there are some scenarios we should take care of,
> 
> 1. User belongs to multiple company, and he has removed from one company. Then he should able to login with the other company.
> 2. If a user belongs to only one company then we can disable the login.
> 
> LMK the thoughts cc: @apoorv1316

@akhilgkrishnan @sudeeptarlekar @apoorv1316 

1. if the user is part of only 1 company has been removed from the company, then the user should not be allowed to login and error message should be displayed on the sign in page when the user tries to login 
2. If the user is part of multiple company and is removed from one company, the user should not be allowed to login in the removed company. But still be logged into the other company automatically
3. if the removed company was default then the other company should be made default automatically. 
4. The company switcher will not be displayed if the user is now part of only one company.

sudeeptarlekar:
I added changes for sign in functionality. 

Regarding the company selection, how user can select the company? Do we have to provide any dropdown on `sign-in` page or do we show a company list where user can select the company in step after sign-in?

apoorv1316 (commented):
@sudeeptarlekar @supriya3105 If a user is in two orgs. and if remove it one from org then the user is not able to login into another org. We get the message "Access denied". Even if we implement a workspace switcher in another PR even the user still needs to be logged in for switching the workspace.

sudeeptarlekar:
@apoorv1316, @supriya3105 any updates on this, its pending since long. Fixed the organization issues as well

PR #653, opened 2022-09-12
Fixed invoice time entries hourly value along with invoice pdf

prasanthchaduvula:
## Notion card
https://www.notion.so/The-qty-in-Invoice-PDF-generated-is-rounded-off-because-of-which-the-invoice-amount-in-miru-and-PDF--d4915e1c12ff4b4b9cac547526f24e06

## Summary

Fixed total hours value in clients page, projects page and team members' total hours and total cost for that particular project. Fixed pdf attachment when client receives the invoice in the mail, hours by the team members getting rounded and because of that, mail view and attached pdf are not matching and the total value is wrong. Fixed this bug.

## Preview

Video: https://www.loom.com/share/c986903b9cf74f179075f28392fd7a73
Video continuation: https://www.loom.com/share/94e125e7387d4de081beb637ee59d540

## Type of change

### Checklist:

### DIFF
--- a/app/javascript/src/components/Clients/Details/index.tsx
+++ b/app/javascript/src/components/Clients/Details/index.tsx
@@ -14,6 +14,7 @@ import AddEditProject from "components/Projects/Modals/AddEditProject";
 import DeleteProject from "components/Projects/Modals/DeleteProject";
 import { cashFormatter } from "helpers/cashFormater";
 import { currencySymbol } from "helpers/currencySymbol";
+import { minutesToHHMM } from "helpers/hhmm-parser";
 import { sendGAPageView } from "utils/googleAnalytics";
 
 import Header from "./Header";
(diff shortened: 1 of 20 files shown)


prasanthchaduvula:
@keshavbiswa As of now I fixed the bugs but we have to create one more issue as the complete hours and price calculation is having duplication of logic. They are getting handled both on the frontend & backend as well. It's not a good idea to have logic for the same feature return on both sides. The current bug also appeared because of that. on the frontend price calculation happens as javascript won't round the value by default but on the backend, ruby rounds the value by default. There are so many components where the logic is repeated in different ways. So it's better to handle these bugs in a different issue. Also, one more thing if you observe the file changes instead of using the existing minutes to hours&minutes helper, developers wrote their own logic and because of that there are many bugs we can see as the minutes to hours&minutes calculation itself is wrong. I made those changes where ever I have seen. It's better to extract components so that others can use them. There are so many places where used badges but it's every time we write that logic in different ways instead of extracting it to component and reusing it which will save so much time for folks and in the day end it helps in sustainable web development

supriya3105:
@prasanthchaduvula Please update the branch.

apoorv1316:
@supriya3105 We are getting the qty now in PDFs. But the qty is more than 60 minutes for entry 5 here also. Because @prasanthchaduvula has fixed the same issue already on the projects page and client page in this PR so it won't be much of an effort to fix it on invoices edit, edit line item, and show invoice page. I can get on a call with him and we can fix this error on all the pages. It would be done by today only
Screenshot 2022-09-13 at 12 55 18 PM

supriya3105:
@apoorv1316 Go ahead and discuss with @prasanthchaduvula . 9.83 should be 10 hours and 23 minutes correct ?

apoorv1316:
> @apoorv1316 Go ahead and discuss with @prasanthchaduvula . 9.83 should be 10 hours and 23 minutes correct ?

Yes, will get it fixed.

prasanthchaduvula:
@supriya3105 @apoorv1316 I resolved all the bugs related to pdf generator, time duration and rate calculation across client, projects and invoices components. This PR only focuses on the above-mentioned things. We can resolve the remaining bugs by creating new issues.

Tomorrow I will extract the total rate calculation to a helper function so that on the frontend we can reuse it instead of rewriting & refactoring each time. In this PR I didn’t because this PR intention is different. 

There are two issues still remaining and Please create below two issues
-> On the invoice generation, manual entries won't work if we are giving quantity in the format of `08:37`. The same thing is also on the present invoice editing page.
-> On invoice generation, while showing the modal for new entries adding please remove search, and also on the edit page to add new time entries update the hour to rate value. Both models are not consistent.

apoorv1316 (approved):
@prasanthchaduvula It looks good now 🚀. Will open separate issues for other related fixes. cc @supriya3105

keshavbiswa (approved):

PR #659, opened 2022-09-14
Fixed manual time entries, editing entries on invoice generation and editing pages

prasanthchaduvula:
## Notion card
https://www.notion.so/6afc9500b2cc42af86be55f37894fb09?v=185bd45b9dc24dfcad9c57c07d32a7c0&p=984c5819652745fc9d8e47771ac184e7&pm=s

## Summary

This PR resolves the bugs in time entries on invoice generation and editing pages. Earlier when user want to add manual entries it was not supporting the time in HH:MM and rate due to that line item total is always wrong or NaN. If user want to edit the entries added on the invoice generation page also the same above mentioned things goes wrong because of that there is no match between line item total, subtotal and amount due.
On the invoice edit page also same bugs are consistent. 

- Now on the invoice generation page with this PR, user can add manual time entries with the HH:MM format as quantity and rate and line total will auto calculated accordingly.
- Now on the invoice generation page with this PR, user can edit the added time entries from new time entries modal. Now the quantity supports HH:MM format as quantity and rate and line total will auto calculated accordingly.

This PR fixes all the bugs related to invoice time entries on both invoice generation & edit page. 
Please watch below videos shared in the preview.

## Preview

https://www.loom.com/share/7901a08e88894bb8913088b3f1339bec

https://www.loom.com/share/62568522c3ca49dca1507e65656ba0b9

## Type of change

### Checklist:

### DIFF
--- a/app/javascript/src/components/Invoices/Generate/ManualEntry.tsx
+++ b/app/javascript/src/components/Invoices/Generate/ManualEntry.tsx
@@ -2,19 +2,21 @@ import React, { useState, useEffect, useRef } from "react";
 
 import { Trash } from "phosphor-react";
 
+import { minutesFromHHMM, minutesToHHMM } from "helpers/hhmm-parser";
+
 const ManualEntry = ({
   entry,
   manualEntryArr,
   setManualEntryArr
 }) => {
-
-  const [name, setName] = useState<string>("");
-  const [date, setDate] = useState<string>("");
-  const [description, setDescription] = useState<string>("");
-  const [rate, setRate] = useState<any>("");
-  const [quantity, setQuantity] = useState<any>("");
-  const [lineTotal, setLineTotal] = useState<any>(entry.lineTotal ? entry.lineTotal : 0);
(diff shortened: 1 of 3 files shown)


apoorv1316:
@prasanthchaduvula Can we also implement the same logic here also to convert this 12:87 to 13:27? Also, can we handle validations too?
Screenshot 2022-09-16 at 11 24 08 AM
Screenshot 2022-09-16 at 11 25 29 AM

prasanthchaduvula:
@apoorv1316  I went through the above comment and quantity on the edit entry is the input box and it’s the user typing it, it’s a live one. If you observe carefully the line total is correct but the quantity input box we are showing what the user enter until he saves it. We can’t instantly change the user input as our UI don’t have any save for individual entry while editing. We are directly using the save button for the complete invoice to update things at one shot. So we can’t change the user quantity input while he is typing and we don’t know whether his typing on the quantity is done or not, until he click on the save button the edit invoice .

apoorv1316 (approved):

swaathi (approved):

PR #661, opened 2022-09-16
Fly deployment added

akhilgkrishnan:
## Type of change

      not work as expected)

### Checklist:


gowsik-ragunath (approved):
LGTM. 

But I don't think this will check if the test suite passes or not if we want to proceed with deployment only after confirming that the test suite passes then can we move this deploy step to the existing workflow CICD.yml?

We can use [jobs.if](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif) to run the deploy step only if the branch is `develop` and [jobs.need](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) to confirm that previous steps ran successfully.

PR #709, opened 2022-10-06
Show latest updated invoices

shalapatil:
## Notion card
https://www.notion.so/saeloun/Update-API-for-invoice-list-page-to-fetch-the-recently-updated-invoices-321b5bfad7e749849ca1af0d3e338089

## Type of change

      not work as expected)

### Checklist:

### DIFF
--- a/app/controllers/internal_api/v1/invoices_controller.rb
+++ b/app/controllers/internal_api/v1/invoices_controller.rb
@@ -16,8 +16,14 @@ class InternalApi::V1::InvoicesController < InternalApi::V1::ApplicationControll
             .order(created_at: :desc),
       items_param: :invoices_per_page)
 
+    recently_updated_invoices = current_company.invoices
+      .includes(:client)
+      .order("updated_at desc")
+      .limit(10)
+
     render :index, locals: {
       invoices:,
+      recently_updated_invoices:,
       pagy: pagy_metadata(pagy),
       summary: current_company.overdue_and_outstanding_and_draft_amount
     }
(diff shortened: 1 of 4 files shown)


app/controllers/internal_api/v1/invoices_controller.rb
```diff
@@ -16,8 +16,14 @@ def index
             .order(created_at: :desc),
       items_param: :invoices_per_page)
 
+    recently_updated_invoices = current_company.invoices
```
prasanthchaduvula, line 19: @shalapatil It's better if we use define scope in invoice model and call it here.
Next thing is here we are limiting the limit to 10 instead it's better if we use `invoices_per_page` param value to have more control and it's better to use the above logic by tweaking it instead of making two requests to database as we are using `.order(created_at: :desc)` in the above query,

prasanthchaduvula (changes requested):

app/controllers/internal_api/v1/invoices_controller.rb
shalapatil: scope as in for what? this is simple query without any where clause. 
also, 10 is the max value which we want to fetch and it is not dependent on page at all. 
Here, regardless of which page's invoices are being shown, we want to show only top 10 recently updated invoices.

supriya3105: @shalapatil Yes. we had agreed to show the 10 updated invoices.

prasanthchaduvula: @shalapatil LGTM

prasanthchaduvula (approved):

prasanthchaduvula:
@apoorv1316 Please test this in local using postman

PR #719, opened 2022-10-07
Elastic search reindexing issue fixed

akhilgkrishnan:
## Summary

In production, we are facing an issue with elastic search reindexing. While backtracing the error, if had found that the issue is because of https://github.com/saeloun/miru-web/blob/df23ba375133e7074acc60695267468cd17bce55/app/models/timesheet_entry.rb#L72 this line. 

Cause of issue: 
This happens only for deleted project timesheet entry reindexing. while fetching the project from the timesheet entry it returns nil and give `undefined client_id for nil class error`

## Type of change

      not work as expected)

### Checklist:

### DIFF
--- a/lib/tasks/fly.rake
+++ b/lib/tasks/fly.rake
@@ -10,7 +10,7 @@ namespace :fly do
   #  - changes to the filesystem made here are DISCARDED
   #  - full access to secrets, databases
   #  - failures here prevent deployment
-  task :release => 'db:migrate:with_data'
+  task :release => 'db:migrate'
 
   # SERVER step:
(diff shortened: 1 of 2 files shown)


lib/tasks/fly.rake
```diff
@@ -10,7 +10,7 @@ namespace :fly do
   #  - changes to the filesystem made here are DISCARDED
   #  - full access to secrets, databases
   #  - failures here prevent deployment
-  task :release => 'db:migrate:with_data'
+  task :release => 'db:migrate'
```
akhilgkrishnan, line 13: Note: This is a fix for staging deployment failure

PR #728, opened 2022-10-12
fix size of current password input field on personal settings page

rajat-tomar:
## Summary
It reduces the size of the current password input field and make it equal to the size of the password input

## Type of change

### Checklist:


rajat-tomar:
@supriya3105 @apoorv1316 why is this check failing?

apoorv1316 (approved):

apoorv1316:
@rajat-tomar Thanks for contributing to miru.

PR #745, opened 2022-10-19
Invoice filter UI update

Shruti-Apte:
## Notion card
https://www.notion.so/saeloun/Refactor-Filter-option-on-Invoice-list-page-as-per-new-design-fe8ad6f314884eeabca33059e2bd4bb5

## Summary

- Updated UI for filter panel
- Outsideclick functionality added
- Applied filters visible on invoice list page
- Few bug fixes
- sidepanel and button (reusable) component added

## Preview

https://www.loom.com/share/0747f23e787c40f4b809dea955f2ef09

https://www.loom.com/share/e467397751314d98b2783bbc3717c11d

## Type of change

      not work as expected)

### Checklist:


app/javascript/src/StyledComponents/SidePanel.tsx
```diff
@@ -0,0 +1,89 @@
...
+          <h4 className={classnames("text-base flex items-center")}>
+            {HeaderLogo}
+            {HeaderTitle}
+          </h4>
+          <button onClick={() => setFilterVisibilty(false)}>
+            <X size={16} />
+          </button>
+        </div>
+      </div>
+      {restprops.children}
```
prasanthchaduvula, line 68: @Shruti-Apte Instead of passing `restprops` Please 
The component should be used in this way to make it more flexible. Above one is not allowing to customize the header and footer
Below one is just an example. We have to allow developer to add and customize the footer as well
```
<SidePanel isOpen={isOpen} onClose={onClose} >
     <SidePanel.Header>
             // This where the developer will write the header component
      </SidePanel.Header>
       <SidePanel.Body>
             // This where the developer will write the body component
         </SidePanel.Body>
         <SidePanel.Footer>
             // This where the developer will write the footer component
         </SidePanel.Footer>
   </SidePanel>
```

app/javascript/src/common/CustomCheckbox.tsx
```diff
@@ -39,7 +42,7 @@ const CustomCheckbox = ({
     </div>
 
     {text !== "" && (
-      <label htmlFor="A3-yes" className="select-none">
+      <label htmlFor={name} className={classnames("select-none",labelClassName)}>
```
prasanthchaduvula, line 45: @Shruti-Apte Why above text is changed to type object, if we are using the string value to appear in label name

app/javascript/src/components/Invoices/List/FilterSideBar/index.tsx
```diff
@@ -63,33 +68,85 @@ const FilterSideBar = ({
...
     }
   };
 
-  const handleSelectFilter = (selectedValue, field) => {
-    if (selectedValue.value === "custom"){
+  const handleSelectFilter = (event) => {
+    const selectedValue = dateRangeOptions.filter(
+      (option) => option.value === event.target.value
+    );
+    if (selectedValue[0].label === "custom") {
```
prasanthchaduvula, line 86: @Shruti-Apte This whole function checking is wrong. What is the label is having upper case text or with some other text along with it. We should always check value which is standard and also here with these changes we are not handling all type of filters. Earlier one is good one to handle all sort of data types either it's an array or object.

app/javascript/src/components/Invoices/List/FilterSideBar/index.tsx
```diff
@@ -63,33 +68,85 @@ const FilterSideBar = ({
...
+        });
     }
-    else {
-      selectedValue.value != "custom" && setFilters({
-        ...filters,
-        [field.name]: selectedValue
-      });
+  };
+
+  const handleCheck = (event) => {
```
prasanthchaduvula, line 101: @Shruti-Apte Can we handle everything in `handleSelectFilter` instead of writing for clients and status separately. Try to make it generic as it will also be useful when we work on reports filter.

app/javascript/src/components/Invoices/List/FilterSideBar/index.tsx
```diff
@@ -144,123 +211,171 @@ const FilterSideBar = ({
...
-        <span className={`${getStatusCssClass(props.data.label)} text-xs tracking-widest`} >
-          {props.data.label}
-        </span>
-      </div>
-    );
-  };
-
-  if (loading){
+  if (loading) {
     return <div>Loading....</div>;
```
prasanthchaduvula, line 221: Here instead of writing loading try to show `miru-logo`. It is useful in production.

app/javascript/src/components/Invoices/List/FilterSideBar/index.tsx
```diff
@@ -144,123 +211,171 @@ const FilterSideBar = ({
...
-              {showCustomFilter &&
-                <div className="mt-1 absolute flex flex-col bg-miru-white-1000 z-20 shadow-c1 rounded-lg">
-                  <CustomDateRangePicker
-                    hideCustomFilter={hideCustomFilter}
-                    handleSelectDate={handleSelectDate}
-                    onClickInput={onClickInput}
-                    selectedInput={selectedInput}
-                    dateRange={dateRange}
+    <SidePanel
+      handleCancel={handleReset}
```
prasanthchaduvula, line 226: @Shruti-Apte Please change the sidepanel props by having header props first, then body props followed by footer props

app/javascript/src/components/Invoices/List/container.tsx
```diff
@@ -13,37 +15,113 @@ const Container = ({
...
-  setInvoiceToDelete
-}) =>
-  invoices.length > 0 ? (
-    <div className="overflow-x-scroll flex flex-col items-stretch">
+  setInvoiceToDelete,
+  filterParams,
+  setFilterParams,
+  filterIntialValues,
+  filterParamsStr
+}) => {
```
prasanthchaduvula, line 23: @Shruti-Apte Please can you post a video explaining why this code is here and what it is for on app.

app/javascript/src/components/Invoices/List/index.tsx
```diff
@@ -19,7 +19,7 @@ import DeleteInvoice from "../popups/DeleteInvoice";
 
 const Invoices: React.FC = () => {
   const filterIntialValues = {
-    dateRange: { label: "All", value: "all", from: "", to: "" },
+    dateRange: [{ label: "All", value: "all", from: "", to: "" }],
```
prasanthchaduvula, line 22: @Shruti-Apte Can I know why dateRange is changed to array.

prasanthchaduvula (changes requested):

app/javascript/src/common/CustomCheckbox.tsx
Shruti-Apte: Passing badge component as a text for status filter, that's why changed type to object

apoorv1316 (changes requested):
@Shruti-Apte After selecting a client in the filter and applying it if we open filters again the checkbox for that client is not showing checked. Can you check that? Also, the overdue status is missing from the status filter options.
https://www.loom.com/share/7ee3fd7a268c43d3a726a8ec6ddeb83b

app/javascript/src/common/CustomCheckbox.tsx
```diff
@@ -8,37 +8,39 @@ const CustomCheckbox = ({
...
+        id={id}
+        name={name}
+        checked={isChecked}
+        onChange={handleCheck}
+        value={checkboxValue}
+        data-cy= 'select-all-checkbox'
+        className="grid col-start-1 row-start-1 w-5 h-5 opacity-0 cursor-pointer custom__checkbox"
+      />
+      <div className="grid col-start-1 row-start-1 justify-items-center content-center flex-shrink-0 w-5 h-5 mr-2 bg-white border-2 border-miru-han-purple-1000 focus-within:border-blue-500">
+        <svg
```
prasanthchaduvula, line 28: @Shruti-Apte instead of svg is it possible to phosphor icons ?

app/javascript/src/components/Invoices/List/FilterSideBar/index.tsx
```diff
@@ -73,79 +81,43 @@ const FilterSideBar = ({
         value: item.id
       }));
       setClientList(clientArr.sort(sortClients));
+      setFilteredClientList(clientArr.sort(sortClients));
```
prasanthchaduvula, line 84: @Shruti-Apte Is individual stateful client list for this component` setFilteredClientList` required ? can't we use the original  client list itself

prasanthchaduvula (changes requested):

app/javascript/src/common/CustomCheckbox.tsx
Shruti-Apte: checkbox is customised with tick to match with the designs. It will take some extra efforts to figure out the same thing with phosphor icons.. I can work on it  if that is a priority !

app/javascript/src/components/Invoices/List/FilterSideBar/index.tsx
Shruti-Apte: I added  `setFilteredClientList` because we need to set the clientlist to original state if no search query is written or clientlist is empty ! So if we use just the clientList state we need to call backend everytime when the query is empty to avoid backend calls I used different state. 
We can connect regarding this !

prasanthchaduvula (approved):

apoorv1316 (approved):

PR #755, opened 2022-10-26
711-Added logic to only update non billed time entries

### LINKED ISSUE #711: Users should not be able to edit and delete a billed time tracking entry.
### Is your proposal related to a problem?

  Currently, employees are able to edit and delete their time entries which are billed which causes confusion in invoicing. 

### Describe the solution you'd like

 Once the time tracking entry of an employee is billed i.e. added to an invoice and sent to client, the employee should not be able to edit or delete that time entry . We need to remove the edit and delete icon for billed time tracking entries. 

The admin and owner role should still be able to edit and delete those entries.

sudeeptarlekar:
Signed-off-by: sudeeptarlekar <sudeeptarlekar@gmail.com>

Fixes #711 

## Summary
Dis-allowed user to update entry after it is billed.

## Type of change

## How Has This Been Tested?
Tested on local on UI.
Tested using test cases.

### Checklist:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   include ErrorHandler
   include CurrentCompanyConcern
   include Pagy::Backend
+  include SetCurrentDetails
 
   before_action :authenticate_user!, :validate_company!
 
--- a/app/javascript/src/components/TimeTracking/EntryCard.tsx
+++ b/app/javascript/src/components/TimeTracking/EntryCard.tsx
@@ -18,6 +18,42 @@ interface props {
   bill_status: string;
 }
 
+const showUpdateAction = (billStatus, id, setEditEntryId) => {
+  if (billStatus != "billed") {
+    return (
+      <button onClick={() => setEditEntryId(id)} className="mx-10">
+        <img
+          src={editIcon}
+          alt="edit"
+          className="icon-hover text-miru-han-purple-600 hover:text-miru-han-purple-1000 w-4 h-4"
+        />
+      </button>
+    );
+  } else {
+    return (
+      <div className="mx-10 w-4 h-4"></div>
+    );
+  }
+};
+
+const showDeleteAction = (billStatus, id, handleDeleteEntry) => {
+  if (billStatus != "billed") {
+    return (
+      <button onClick={() => handleDeleteEntry(id)} className="mr-10">
+        <img
+          src={deleteIcon}
+          alt="delete"
+          className="icon-hover fill-blue text-miru-han-purple-1000 hover:text-miru-han-purple-1000 w-4 h-4"
+        />
+      </button>
+    );
+  } else {
+    return (
+      <div className="mr-10 w-4 h-4"></div>
+    );
+  }
+};
+
 const EntryCard: React.FC<props> = ({
   id,
   client,
@@ -63,20 +99,8 @@ const EntryCard: React.FC<props> = ({
         />
       )}
       <p className="text-4xl ml-6">{minToHHMM(duration)}</p>
-      <button onClick={() => setEditEntryId(id)} className="mx-10">
-        <img
-          src={editIcon}
-          alt="edit"
-          className="icon-hover text-miru-han-purple-600 hover:text-miru-han-purple-1000 w-4 h-4"
-        />
-      </button>
-      <button onClick={() => handleDeleteEntry(id)} className="mr-10">
-        <img
-          src={deleteIcon}
-          alt="delete"
-          className="icon-hover fill-blue text-miru-han-purple-1000 hover:text-miru-han-purple-1000 w-4 h-4"
-        />
-      </button>
+      { showUpdateAction(bill_status, id, setEditEntryId) }
+      { showDeleteAction(bill_status, id, handleDeleteEntry) }
     </div>
   </div>
 );
--- a/app/javascript/src/components/TimeTracking/EntryCard.tsx
+++ b/app/javascript/src/components/TimeTracking/EntryCard.tsx
@@ -4,6 +4,8 @@ import React from "react";
 import { minToHHMM } from "helpers";
 import { Badge } from "StyledComponents";
 
+import { Roles } from "../../constants";
+
 const deleteIcon = require("../../../../assets/images/delete.svg");
 const editIcon = require("../../../../assets/images/edit.svg");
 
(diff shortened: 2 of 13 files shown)


app/javascript/src/components/TimeTracking/EntryCard.tsx
```diff
@@ -63,20 +99,8 @@ const EntryCard: React.FC<props> = ({
...
-        />
-      </button>
-      <button onClick={() => handleDeleteEntry(id)} className="mr-10">
-        <img
-          src={deleteIcon}
-          alt="delete"
-          className="icon-hover fill-blue text-miru-han-purple-1000 hover:text-miru-han-purple-1000 w-4 h-4"
-        />
-      </button>
+      { showUpdateAction(bill_status, id, setEditEntryId) }
```
prasanthchaduvula, line 102: @sudeeptarlekar Instead of defining view functions why can't we use 
```
billStatus != "billed" && 
<button onClick={() => handleDeleteEntry(id)} className="mr-10">
        <img
          src={deleteIcon}
          alt="delete"
          className="icon-hover fill-blue text-miru-han-purple-1000 hover:text-miru-han-purple-1000 w-4 h-4"
        />
      </button>
```

prasanthchaduvula: @sudeeptarlekar Here we also have to check The admin and owner role should still be able to edit and delete those entries.

app/javascript/src/components/TimeTracking/EntryCard.tsx
```diff
@@ -63,20 +99,8 @@ const EntryCard: React.FC<props> = ({
...
+      { showDeleteAction(bill_status, id, handleDeleteEntry) }
```
prasanthchaduvula, line 103: @sudeeptarlekar Here also why can't we use 
```
billStatus != "billed" && 
<button onClick={() => handleDeleteEntry(id)} className="mr-10">
        <img
          src={deleteIcon}
          alt="delete"
          className="icon-hover fill-blue text-miru-han-purple-1000 hover:text-miru-han-purple-1000 w-4 h-4"
        />
      </button>
```
Here we also have to check The admin and owner role should still be able to edit and delete those entries.

prasanthchaduvula (changes requested):

app/javascript/src/components/TimeTracking/EntryCard.tsx
sudeeptarlekar: We can, but imo its better to keep render function as sleek as possible. Adding suggested code will hamper readability.
Thoughts?

sudeeptarlekar: Also if time entry is non editable we have to render empty div to keep UI consistent.

prasanthchaduvula: @sudeeptarlekar Can't we fix the div UI as it is not needed and regarding the render function please add these checks The admin and owner role should still be able to edit and delete those entries.

sudeeptarlekar: But from where admin/owner can change time entries of employees? And M not sure which div UI is not needed

prasanthchaduvula: @sudeeptarlekar The <div> in else condition of the ternary.

sudeeptarlekar: That is needed div

apoorv1316 (changes requested):
@sudeeptarlekar As per the card description, the admin and owner should still be able to delete and edit the billed entries but the edit and delete icons are missing for the owner and admin roles. https://www.loom.com/share/da2da1f1b4f7424d8965336918c9fbf8

supriya3105:
@apoorv1316 Please test this and @prasanthchaduvula please review again.

apoorv1316 (approved):

app/controllers/application_controller.rb
```diff
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   include ErrorHandler
   include CurrentCompanyConcern
   include Pagy::Backend
+  include SetCurrentDetails
```
prasanthchaduvula, line 9: @sudeeptarlekar Can we rename it to `SetCurrentUserDetails`

prasanthchaduvula: @sudeeptarlekar Can I know what is the reason for `SetCurrentDetails` as we already have the access to `current user, current company and user roles`

app/models/current.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Current < ActiveSupport::CurrentAttributes
```
prasanthchaduvula, line 3: @sudeeptarlekar No need of this model. You can check it from the controller before updating or deleting the timesheet entry.

app/policies/timesheet_entry_policy.rb
```diff
@@ -14,7 +14,7 @@ def create?
   end
 
   def update?
-    record.user_id == user.id ||
+    (record.user_id == user.id && !record.billed?) ||
```
prasanthchaduvula, line 17: @sudeeptarlekar Here any admin or owner from the other companies can hamper our data as we are only checking the user role.

Here is an example
```
def update?
    authorize_current_user
  end

private
def authorize_current_user
    unless user.current_workspace_id == record.company_id
      @error_message_key = :different_workspace
      return false
    end

  end
```

prasanthchaduvula (changes requested):

app/controllers/application_controller.rb
sudeeptarlekar: You only have access to these objects in conteollers

prasanthchaduvula: @sudeeptarlekar if you need `current_user` and `current_company` I don't think you need this `SetCurrentDetails`. They are already available to controllers

prasanthchaduvula: @sudeeptarlekar Can you write what we are going to get with this `SetCurrentDetails`

sudeeptarlekar: We need current_user in model which is not available, this module will make it available in models as well!!!

sudeeptarlekar: Instead of writing I would like you to try it on local, and if you have better solution that this, please enlighten me

app/policies/timesheet_entry_policy.rb
sudeeptarlekar: Are you crazy?? we are checking if user has a role for company or not

sudeeptarlekar: FYI, we have test cases for same as well
PLEASE check it [here](https://github.com/saeloun/miru-web/blob/711-user-should-not-be-able-to-update-billed-entries/spec/policies/timesheet_entry_policy_spec.rb#L98)

app/controllers/application_controller.rb
prasanthchaduvula: @sudeeptarlekar https://www.loom.com/share/a7e0c79b44704f98bb39ebc364a6ea48

sudeeptarlekar: What you are suggesting is not best practice either, but here I will ask team to take a call.

prasanthchaduvula (approved):

PR #788, opened 2022-11-20
Some Cleanups

vipulnsward:
What:
- Fix Gemfile
- Remove dead files


shalapatil (approved):
LGTM, just a question - why did we make simple utility fn as a service?

PR #807, opened 2022-11-29
Add and setup parallel_spec for faster local testing

vipulnsward:
@prasanthchaduvula can you test this locally? Reduces run time of complete test suite to 40 seconds for me.

Needs some exceptions to look at


prasanthchaduvula:
@vipulnsward I ran the tests parallelly. I didn't notice much difference. Every time the test running time got varied still it is not noticeable. Please let me know.

PR #821, opened 2022-12-02
Resolved date and day mistmatch issue

sudeeptarlekar:
PR #818 has few bugs that when you click on any date in the month, date will not get set correctly. Reverted few changes in old PR and hard coded the days for month.

### Preview
https://user-images.githubusercontent.com/23317691/205284081-63643a6f-ffd1-4d6e-b1c1-68a19965ba5d.mp4

### DIFF
--- a/app/javascript/src/components/TimeTracking/MonthCalender.tsx
+++ b/app/javascript/src/components/TimeTracking/MonthCalender.tsx
@@ -12,7 +12,6 @@ dayjs.Ls.en.weekStart = 1;
 const MonthCalender = ({
   fetchEntries,
   selectedEmployeeId,
-  dayInfo,
   entryList,
   selectedFullDate,
   setSelectedFullDate,
@@ -287,7 +286,6 @@ const MonthCalender = ({
 interface Iprops {
   fetchEntries: (from: string, to: string) => void;
   selectedEmployeeId: number;
-  dayInfo: any[];
   selectedFullDate: string;
   setSelectedFullDate: any;
   entryList: object;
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/TimeTracking/MonthCalender.tsx
```diff
@@ -212,12 +212,12 @@ const MonthCalender = ({
       </div>
       <div className="bg-miru-gray-100 p-4">
         <div className="mb-4 flex justify-between bg-miru-gray-100">
-          {dayInfo.map((d, index) => (
+          {["Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"].map((d, index) => (
```
prasanthchaduvula, line 215: @sudeeptarlekar if `dayInfo ` is an unused prop & is not used anymore, please remove it even from the time-tracking module as well. One more thing is `dayinfo` is getting used to show the days information in the week view and in the day view as well. Instead of hardcoding the complete day names, it's better to reuse them.

prasanthchaduvula (changes requested):

supriya3105:
@sudeeptarlekar Please resolve the failing tests on this PR so that we can merge it.

sudeeptarlekar:
@supriya3105 fixed specs

app/javascript/src/components/TimeTracking/MonthCalender.tsx
sudeeptarlekar: `dayInfo` is using weekDay for calculating the days in the month, but month view imo should not change days displayed on top based on present day, so using `dayInfo` for rendering days is not ideal, In case of week's display it does make sense as we are only showing the week and it does not matter from which day week is starting in week view. If we want to show week from Monday irrespective of current day then we should hardcode days in week view as well.

apoorv1316:
> Specs still failing? Can you please recheck?

Shalaka has opened a PR for fixing this spec. It's failing on multiple PR's

prasanthchaduvula (approved):

PR #894, opened 2022-12-29
Add Discard Model checks on Invoices table

vipulnsward:
What:
- Add Discard Model checks on Invoices table

Why:
- Invoices data should never be deleted

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -49,3 +49,7 @@ ENABLE_WEEKLY_REMINDER=
 
 # Sentry
 SENTRY_DSN=
+
+#Database
+DB_USER=
+DB_PASS=
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,7 +3,8 @@ default: &default
   encoding: unicode
   host: localhost
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  password:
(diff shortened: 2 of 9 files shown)


prasanthchaduvula:
@vipulnsward Please review it

vipulnsward (commented):
Are specs added to all places where kept is used?

prasanthchaduvula:
@vipulnsward specs added to all places where kept is used

config/database.yml
```diff
@@ -3,15 +3,16 @@ default: &default
   encoding: unicode
   host: localhost
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  password:
+  user: <%= ENV.fetch("DB_USER") { "" } %>
```
vipulnsward, line 6: Need to add `DB_USER`, `DB_PASS` to be added to sample env

prasanthchaduvula: @vipulnsward Done

.env.example
```diff
@@ -49,3 +49,7 @@ ENABLE_WEEKLY_REMINDER=
 
 # Sentry
 SENTRY_DSN=
+
+#Database
+DB_USER=
+DB_PASS=
```
vipulnsward, line 55: @prasanthchaduvula please fix your editor to avoid these EOF issues

PR #948, opened 2023-01-16
Hide Billing tab under settings for admin and owner user role

sanjibroy360:
## Notion card
[https://www.notion.so/saeloun/Hide-Billing-tab-under-settings-for-admin-and-owner-user-role-3bf1fbbef94045b7a45797dc4fd2bdda](https://www.notion.so/saeloun/Hide-Billing-tab-under-settings-for-admin-and-owner-user-role-3bf1fbbef94045b7a45797dc4fd2bdda)

## Summary
Since the billing tab is currently not developed, hence hide the Billings tab for admin and owner user roles. Here inside the `SideNav` component (`app/javascript/src/components/Profile/SubNav.tsx`) if the value of `isAdmin` is `true` I'll hide the `Billing` tab.

## Preview
[https://www.loom.com/share/d262c9a7a22e42b5b98f1574b1c0dd3b](https://www.loom.com/share/d262c9a7a22e42b5b98f1574b1c0dd3b)

## Type of change
-  In the `SideNav` component (`app/javascript/src/components/Profile/SubNav.tsx`) if the value of `isAdmin` is `true` I'll hide the "Billing" tab.

- [ ] Refactor (non-breaking and retains same functionality)
      not work as expected)

## How Has This Been Tested?

1. Log in as an admin or owner user.
2. Click on setting from the left navigation pane.
3. It should not show the "Billing" tab under the Sidenav bar of the Settings page.

### Checklist:

- [x] I have annotated changes in the PR that are relevant to the reviewer

### DIFF
--- a/app/javascript/src/components/Profile/SubNav.tsx
+++ b/app/javascript/src/components/Profile/SubNav.tsx
@@ -53,15 +53,17 @@ const SideNav = ({ isAdmin, firstName, company, lastName, email }) => {
           PAYMENT SETTINGS
         </NavLink>
       </li>
-      <li className="border-b-2 border-miru-gray-400">
-        <NavLink
-          end
-          className={({ isActive }) => getActiveClassName(isActive)}
-          to="/profile/edit/billing"
-        >
-          BILLING
-        </NavLink>
-      </li>
+      {!isAdmin && (
+        <li className="border-b-2 border-miru-gray-400">
+          <NavLink
+            end
+            className={({ isActive }) => getActiveClassName(isActive)}
+            to="/profile/edit/billing"
+          >
+            BILLING
+          </NavLink>
+        </li>
+      )}
       <li className="border-b-2 border-miru-gray-400">
         {/* <NavLink end to="/profile/edit/import" className={({ isActive }) => getActiveClassName(isActive)}>
           IMPORT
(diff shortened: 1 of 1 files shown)


app/javascript/src/components/Profile/SubNav.tsx
```diff
@@ -53,15 +53,17 @@ const SideNav = ({ isAdmin, firstName, company, lastName, email }) => {
...
-      <li className="border-b-2 border-miru-gray-400">
-        <NavLink
-          end
-          className={({ isActive }) => getActiveClassName(isActive)}
-          to="/profile/edit/billing"
-        >
-          BILLING
-        </NavLink>
-      </li>
+      {!isAdmin && (
```
onkar-saeloun, line 56: Why do we need !isAdmin flag?
I think you should comment it if you are not using it currently.

sanjibroy360: As the requirement was to hide that tab only for two specific roles that's why using the `isAdmin` flag I hid that Billing tab only for those two roles(admin, owner).

onkar-saeloun: Actually the wrapper function itself is for admin so I don't see any point to add such condition.

PR #968, opened 2023-01-24
Time tracking specs cypress

apoorv1316:
## Summary
1. Admin 

- should select a date and add an entry
- should be able to edit an entry
- should be able to delete a timesheet entry
- should be able to view other user's entries

2. Employee

- should select a date and add an entry
- should be able to edit an entry
- should be able to delete a timesheet entry


supriya3105:
@apoorv1316 Can you add the list of time tracking cypress specs that are part of this PR.

cypress/e2e/timetracking/admin.spec.js
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(function () {
+    cy.visit(signInPath);
+    cy.fixture("credentials").then(function (data) {
+      this.data = data;
+    });
+    cy.loginAsAdmin();
+    cy.visit(timetracking);
+  });
+
+  it("should select a date and add entry", function () {
```
mohinid, line 17: hey, we are writing the same steps for both admin and employee . Can you please check if there is a way it can be refactored?

mohinid (changes requested):
Let's discuss on monday for refactoring.

supriya3105:
@apoorv1316 Closing this as per Vipul's feedback to move the cypress specs in another repository. 
 https://saeloun.slack.com/archives/C03U02LUKSP/p1675611122594079

PR #992, opened 2023-02-03
Fix Seeds data to always use US information

vipulnsward:
What:

- Fix Seeds data to always use US information

Why:

- Our clients are primarily in US so we want to make sure things work well for US first


supriya3105:
@prasanthchaduvula Please verify after the change everything is working fine and merge.

PR #998, opened 2023-02-06
Invoice UX Feedback - Add entry to the list if name field is present 

Shruti-Apte:
### **What :**
If the user enter only the name and clicks outside then the entry should be added in the list 
(First task from this ticket https://www.notion.so/saeloun/Invoice-page-UX-feedback-c94100068cbf41708986ccccfdb92f08)

### **Why :**
Currently  (name and date ) both the fields are mandatory to add the entry.
Requirement: Only name is mandatory to add the entry in the list.

### **Preview :**
**Code:**
```
const handleAddEntry = () => {
    if (!lineItem.date) {
      lineItem.date = new Date().toISOString();
    }

    if (!addNew && lineItem.name) {
      setManualEntryArr([...manualEntryArr, lineItem]);
    }
  };
```
**Screenshots:**
Screenshot 2023-02-06 at 2 27 05 PM

**Video:**
https://www.loom.com/share/5c3101a59b11460488b1da9d278f51ee

### **Todos** (for testing):
Check if manual entry is getting added in the list if name field is filled.


supriya3105:
@Shruti-Apte We don't need to add today's date as default date . Please remove it ? 
The user should just be able to save the entry with only the name as well.

apoorv1316:
@supriya3105 So, if a manual entry is added with just the name, what would be that entry's date?

Shruti-Apte:
> @Shruti-Apte We don't need to add today's date as default date . Please remove it ? The user should just be able to save the entry with only the name as well

@supriya3105 We have backend validations for date & quantity  that’s why added today's date as default date.

supriya3105:
> > @Shruti-Apte We don't need to add today's date as default date . Please remove it ? The user should just be able to save the entry with only the name as well
> 
> @supriya3105 We have backend validations for date & quantity that’s why added today's date as default date.

adding a new ticket to get this fixed from backend. Approving this for now.

supriya3105 (approved):

prasanthchaduvula (approved):

apoorv1316:
@supriya3105 @Shruti-Apte If we want to save a manual with just the name then we have to remove this validation on saving and sending invoice
Screenshot 2023-02-07 at 4 08 17 PM

apoorv1316 (approved):

PR #963, opened 2023-01-19
Added signup, signin, forgot password, reset password API and react components

prasanthchaduvula:
## Notion card
https://www.notion.so/saeloun/Redevelop-Sign-In-page-using-React-1ccc0724feb24e7485a02bdf4c51a17a
https://www.notion.so/saeloun/Redevelop-Sign-Up-page-using-React-58ea46ef7bbd4b239b191ca1863d1063

## Summary

- Created authentication with react components for a signup, signing, forgot password and reset password.
- Created API endpoints for the complete auth flow and integrated them with the frontend.

## Preview
[https://share.vidyard.com/watch/hjA2H2MpCDzcDCSqjsFw3v](https://share.vidyard.com/watch/hjA2H2MpCDzcDCSqjsFw3v)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactor (non-breaking and retains same functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
      not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [ ] I have manually tested all workflows
- [ ] I have performed a self-review of my own code
- [ ] I have annotated changes in the PR that are relevant to the reviewer
- [ ] I have added automated tests for my code

PR #1001, opened 2023-02-06
authenticate user using auth_token

prasanthchaduvula:
### What
- Added `authenticate_user_using_x_auth_token` authenticable concern for json requests to access the internal apis.

### Why
- It is needed for our desktop, web app and even from postman we can simply send a request with token and access the data. It makes easier to send request to our API.


prasanthchaduvula:
@ajinkyaa Once you log in to the Miru web app on your local computer. Please pass `X-Auth-Email `and `X-Auth-Token` in the request headers whenever you want to access the internal API.
Here `X-Auth-Email` is the user email and `X-Auth-Token` is the user token.

supriya3105:
Closing this as it is already in the #963

PR #1002, opened 2023-02-07
Fix team controller index view

shalapatil:
Notion link
https://www.notion.so/saeloun/Refactor-controllers-51ad6bf864c448308db7a9c6f65241ad

There were helpers defined in the view. Refactored the code to move the code to presenter class.


app/presenters/team_presenter.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  private
+
+    def formatted_team
+      team.map do |company_user|
+          member = company_user.user
+          {
+            id: member.id,
+            # profile_picture: user_avatar(member),
```
vipulnsward, line 27: Is this not needed? If yes, lets remove

app/presenters/team_presenter.rb
```diff
@@ -0,0 +1,71 @@
...
+            status: team_member_status(member),
+            member:
+          }
+        end
+    end
+
+    def formatted_invitations
+      invitations.map do |member|
+          {
+            # profile_picture: image_url "avatar.svg",
```
vipulnsward, line 42: ditto

app/presenters/team_presenter.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+class TeamPresenter
```
vipulnsward, line 3: We try to keep the scope limited per presenter, and name it accordingly. 
So this can be called `TeamIndexPresenter`, then you can simply call `#process` on it

app/presenters/team_presenter.rb
```diff
@@ -0,0 +1,71 @@
...
+          {
+            id: member.id,
+            # profile_picture: user_avatar(member),
+            first_name: member.first_name,
+            last_name: member.last_name,
+            name: member.full_name,
+            email: member.email,
+            role: member.primary_role(current_company),
+            status: team_member_status(member),
+            member:
```
vipulnsward, line 34: If we can avoid passing this pass profile url from that, that would be great, not a big deal though.

vipulnsward (approved):
lgtm, overall, please address the comments and merge

app/presenters/team_presenter.rb
shalapatil: Okay, not sure what's the standard practice here. 
We have added multiple presenters but for like model like for singular and list ex, client_presenter and clients_presenter. There also we have defined methods as such it is extensible for different views tomorrow. 
what should we follow here?

app/presenters/team_presenter.rb
shalapatil: To do that, 
we need to include multiple view helpers in the presenter 
also, need to add `Rails.application.routes.default_url_options[:host]` in all env files.
Should I make that change?

vipulnsward: No, its okay for now

app/presenters/team_presenter.rb
vipulnsward: Aah, okay. Let me refactor a bit and we can follow those conventions.

PR #1009, opened 2023-02-08
Chore: Run a bundle update

vipulnsward:
What:  Run a bundle update

Why: Check for breaking gems


vipulnsward:
@apoorv1316 please verify things are working fine locally.

apoorv1316:
@vipulnsward Yes it is working fine locally.

PR #1012, opened 2023-02-08
Manual line item entry with 0 quantity

shalapatil:
Notion: https://www.notion.so/saeloun/Refactor-the-validations-added-to-invoice-page-while-adding-line-items-1-47988406ad394094b63c0581bda92dcd

What: We needed a functionality to add manual line item entry without any quantity

Why: Our code had validation to check if quantity is greater than 0, removed that check


supriya3105:
Closing this as complete requirement was not provided.

PR #1018, opened 2023-02-09
Invoice filter bug

onkar-saeloun:
What-
PR includes the change in 
1. The custom date picker component which is a common component. (Now onwards the component will behave such as after selecting From date field the focus will be automatically directed to To  date field ).
2. Invoice filter UI has an issue that whenever there will be more clients the Apply and Reset buttons were not visible. Also, the user was not been able to scroll it now user can scroll in a worst-case scenario.
3. Invoice filter can be closed by pressing `esc` key on keyboard(used window event to achieve this.). The custom date picker will be closed by an outside click(using a common hook to achieve this).

Why-
- The change in the custom date picker will improve user experience in all the places where it is user.
- Filter panel issue will be solved in the worst case scenario.

Notion Card-
https://www.notion.so/saeloun/Invoice-list-page-Filter-panel-bugs-87116f47416f451b81f71bc2f074bbae

Loom Video-
https://www.loom.com/share/369f85c95c004313939f801976eff684

Impacting Areas-
- All the custom date picker components
- Filter Sidebar on the Invoice Page

Tested Scenarios-

- [x] Tested sidebar by adding 20 client records.
- [x] Tested Custom date picker on Reports page as well to confirm changes are not breaking


app/javascript/src/components/Invoices/List/FilterSideBar/index.tsx
```diff
@@ -227,244 +230,255 @@ const FilterSideBar = ({
...
-                {isDateRangeOpen ? (
-                  <MinusIcon size={16} />
-                ) : (
-                  <PlusIcon size={16} />
-                )}
+    <SidePanel
+      WrapperClassname="z-50 justify-content-between"
+      setFilterVisibilty={setIsFilterVisible}
+    >
+      <div>
```
onkar-saeloun, line 237: Wrapped <Sidepane.Header> and  <SidePanel.Body> with <div> that's why showing indentation changes.

prasanthchaduvula (approved):

PR #1029, opened 2023-02-11
Basic setup for feature specs

vipulnsward:
What:
- Basic setup for feature specs
- Add example login spec

Why:
- We want to cover more integration specs in future


gowsik-ragunath:
@vipulnsward CI issue is fixed.

Gemfile
```diff
@@ -201,7 +201,10 @@ group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem "capybara", ">= 3.26"
   gem "selenium-webdriver", ">= 4.0.0"
-  gem "webdrivers"
+  # Ref: https://www.plymouthsoftware.com/articles/rails-on-docker-system-specs-in-containers-with-rspec-capybara-chrome-and-selenium/
+  unless ENV["CI"]
```
vipulnsward, line 205: Is there are better way to achieve this?

spec/spec_helper.rb
```diff
@@ -39,10 +39,13 @@
   config.verbose_retry = true
   config.display_try_failure_messages = true
 
-  # run retry only on features
-  config.around do |ex|
-    ex.run_with_retry retry: 3
-  end
+  # config.around do |example|
```
vipulnsward, line 42: In a follow up PR lets uncomment this. Then lets add this:
- Screenshot when capybara spec fails

spec/rails_helper.rb
```diff
@@ -54,16 +55,31 @@
...
     Faker::UniqueGenerator.clear
     OmniAuth.config.test_mode = true
   end
 
   config.around do |example|
-    if example.metadata[:feature]
-      VCR.turned_off { example.run }
-    else
-      example.run
+    if [:system, :feature].include?(example.metadata[:type])
```
vipulnsward, line 66: 👍

spec/support/capybara.rb
```diff
@@ -0,0 +1,30 @@
...
+Capybara.server = :puma
+
+if ENV["CI"].present?
+  Capybara.register_driver :chrome_headless do |app|
+    options = Selenium::WebDriver::Chrome::Options.new(args: %w[headless window-size=1400,1000])
+    options.add_argument(
+      "--enable-features=NetworkService,NetworkServiceInProcess",
+    )
+
+    if ENV["HUB_URL"]
```
vipulnsward, line 15: We should avoid `HUB_URL`  since it can indicate need to setup or use a service like Browserstack, etc. We can keep it for now, but can add a follow up task to fix this.

spec/support/database_cleaner.rb
```diff
@@ -12,11 +12,24 @@
     DatabaseCleaner.strategy = :transaction
   end
 
+  config.before(:each, type: :system) do
+    # :rack_test driver's Rack app under test shares database connection
+    # with the specs, so continue to use transaction strategy for speed.
+    driver_shares_db_connection_with_specs = Capybara.current_driver == :rack_test
```
vipulnsward, line 18: Are we using  `rack_test`anywhere?

spec/feature/users/login_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "User Sign In", type: :system do
```
vipulnsward, line 5: type  == system here, but folder is feature since I had added it that way. Lets move the spec to `system` folder.

docker-compose.test.yml
```diff
@@ -60,24 +68,6 @@ services:
     depends_on:
       - database
 
-  sidekiq:
```
vipulnsward, line 63: 👍

.rubocop.yml
```diff
@@ -513,6 +513,7 @@ RSpec/DescribeClass:
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribeClass
   Exclude:
     - "spec/requests/**/*"
+    - "spec/feature/**/*"
```
vipulnsward, line 516: Need to change to "/system/" or add another line for it, else linters will fail.

vipulnsward (commented):
@gowsik-ragunath added some comments.

spec/support/database_cleaner.rb
gowsik-ragunath: No, we are not using rack_test, will remove.

Gemfile
gowsik-ragunath: Not sure, but will check if there is an alternative to add `gem "webdrivers"` in local environment.

vipulnsward: We can do in followup so others are unblocked on this PR

spec/spec_helper.rb
gowsik-ragunath: We are already saving screenshots when the capybara spec fails but in CI I don't think we can access it as this files is saved in the instance,

Screenshot 2023-02-16 at 4 06 57 PM

vipulnsward: Cool, lets do this in followup to save on Buildkite and make it accessible on checking CI failures

.rubocop.yml
gowsik-ragunath: Added system

spec/feature/users/login_spec.rb
gowsik-ragunath: Moved the spec to system folder

spec/support/database_cleaner.rb
gowsik-ragunath: Removed rack_test

gowsik-ragunath:
@vipulnsward I have addressed your comments.

PR #1030, opened 2023-02-13
Fix project access on time tracking

shalapatil:
Notion: https://www.notion.so/saeloun/Unable-to-remove-a-team-member-from-a-project-345c86358a1c443da44369814ace6ef5?pvs=4

Why: Even when member was removed from project, he was getting that project name listed on time tracking page. 

What: Fixed the function which returns the list of user project by considering project members so that user won't see any project from which he is removed


app/models/user.rb
```diff
@@ -91,6 +91,28 @@ class User < ApplicationRecord
   after_discard :discard_project_members
   before_create :set_token
 
+  def allowed_clients(company)
```
vipulnsward, line 94:
```suggestion
  def allowed_clients_for(company)
```

app/models/user.rb
```diff
@@ -91,6 +91,28 @@ class User < ApplicationRecord
...
+    if self.has_admin_access(company)
```
vipulnsward, line 95: Here `self` is redundant

app/models/user.rb
```diff
@@ -91,6 +91,28 @@ class User < ApplicationRecord
...
+    else
+      clients.kept
+        .where(company_id: company.id)
+        .merge(project_members.kept)
+        .order(:name)
+    end
+  end
+
+  def allowed_projects(company)
+    if self.has_admin_access(company)
```
vipulnsward, line 106: Ditto

app/models/user.rb
```diff
@@ -91,6 +91,28 @@ class User < ApplicationRecord
...
+      company.clients.kept.order(:name)
+    else
+      clients.kept
+        .where(company_id: company.id)
+        .merge(project_members.kept)
+        .order(:name)
+    end
+  end
+
+  def allowed_projects(company)
```
vipulnsward, line 105:
```suggestion
  def allowed_projects_for(company)
```

app/services/time_tracking_index_service.rb
```diff
@@ -56,19 +56,13 @@ def set_is_admin
...
+      @clients = current_user.allowed_clients(current_company)
+        .includes(:projects)
     end
 
     def set_projects
-      @projects = Client::IndexPresenter
-        .new(clients, current_company, current_user)
-        .projects_grouped_by_client_name
+      @projects = {}
+      user_projects = current_user.allowed_projects(current_company)
```
vipulnsward, line 65: @shalapatil I think what we are trying to do here should be done via Pundit. Please check out https://github.com/varvet/pundit#scopes and give it a try, instead of adding the methods on User class

app/services/time_tracking_index_service.rb
```diff
@@ -56,19 +56,13 @@ def set_is_admin
...
     def set_clients
-      if is_admin
-        @clients = current_company.clients.kept.order(name: :asc).includes(:projects)
-      else
-        @clients = current_user.clients.kept
-          .where(company_id: current_company.id)
-          .order(name: :asc)
-          .includes(:projects).distinct
-      end
+      @clients = current_user.allowed_clients(current_company)
```
vipulnsward, line 59: Same here, we should use https://github.com/varvet/pundit#scopes I think?

vipulnsward (commented):
@shalapatil added some comments

app/services/time_tracking_index_service.rb
shalapatil: Used pundit scopes

app/services/time_tracking_index_service.rb
shalapatil: Used pundit scopes

app/models/user.rb
shalapatil: removed this

app/models/user.rb
shalapatil: removed this

app/models/user.rb
shalapatil: removed this

app/models/user.rb
shalapatil: removed this

app/models/user.rb
```diff
@@ -108,6 +108,10 @@ def full_name
     "#{first_name} #{last_name}"
   end
 
+  def has_admin_access(company)
```
vipulnsward, line 111: This should be in scope too

supriya3105:
@shalapatil Please check why the build is failing on this PR and then assign it to Vipul for review.

PR #1038, opened 2023-02-15
Fix allowed_urls value

shalapatil:
What: Sets the allowed urls to ENV["DATABASE_URL"]

Why: Specs were failing locally as database cleaning was failing due to hardcoded sets of database urls. Set that to value from ENV


gowsik-ragunath (approved):
LGTM

PR #1045, opened 2023-02-16
Only index kept records in ES

vipulnsward:
What:
- Only index kept Invoice records in ES

Why:
- So that discarded once don't show up accidentally


vipulnsward:
@shalapatil can you add specs for this?

@supriya3105 any concerns?

PR #1052, opened 2023-02-18
Return only kept invoices data in summary

vipulnsward:
What:
- Return only kept invoices data in summary

Why:
- So that we don't return summary from deleted invoices


vipulnsward:
Merging to fix CI, @shalapatil  / @swaathi please do a post merge review.

PR #1063, opened 2023-02-21
Fixed flaky specs

prasanthchaduvula:
### Why
The two below specs are flaky which makes buildkite fail
- `./spec/models/client_spec.rb:150`
- `./spec/models/invoice_spec.rb:76`

### What
- With this, we fixed the above two failing specs. 
- Updated calculation of  outstanding_amount and fixed`./spec/models/client_spec.rb:150` 
- Added draft, paid, and overdue invoice variables to calculate the statuses invoices which are hardcoded in the code and fixed `./spec/models/invoice_spec.rb:76`


spec/models/client_spec.rb
```diff
@@ -151,8 +151,7 @@
         currency = client.company.base_currency
         status_and_amount = client.invoices.group(:status).sum(:amount)
         status_and_amount.default = 0
-        outstanding_amount = status_and_amount["sent"] + status_and_amount["viewed"]
-        + status_and_amount["overdue"]
+        outstanding_amount = status_and_amount["overdue"]
```
shalapatil, line 154: this is wrong, please fix.
I have made fix for this here. https://github.com/saeloun/miru-web/pull/1030/files

vipulnsward:
Verified locally

PR #1071, opened 2023-02-23
Added auth context and its reducer to handle authentication on client side

prasanthchaduvula:
Closes

Why

What
This PR is part of the feature:move-authentication to react. In this PR we created auth context and its reducer to handle the client-side routing.

### DIFF
--- a/app/javascript/src/utils/storage.ts
+++ b/app/javascript/src/utils/storage.ts
@@ -1,0 +1,27 @@
+import Logger from "js-logger";
+
+const setToLocalStorage = (key, value) => {
+  if (value !== null) {
+    localStorage.setItem(key, JSON.stringify(value));
+  } else localStorage.removeItem(key);
+};
+
+const getFromLocalStorage = key => {
+  let response = "";
+  try {
+    const value = localStorage.getItem(key);
+    response = value ? JSON.parse(value) : "";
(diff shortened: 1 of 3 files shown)


app/javascript/src/utils/storage.ts
```diff
@@ -0,0 +1,27 @@
...
+    response = value ? JSON.parse(value) : "";
+  } catch (error) {
+    Logger.error(error);
+    response = "";
+  }
+
+  return response;
+};
+
+const clearLocalStorageCredentials = () => {
```
vipulnsward, line 22:
```suggestion
const clearCredentialsFromLocalStorage = () => {
```

app/javascript/src/utils/storage.ts
```diff
@@ -0,0 +1,27 @@
+import Logger from "js-logger";
+
+const setToLocalStorage = (key, value) => {
+  if (value !== null) {
+    localStorage.setItem(key, JSON.stringify(value));
+  } else localStorage.removeItem(key);
+};
+
+const getFromLocalStorage = key => {
```
vipulnsward, line 9:
```suggestion
const getValueFromLocalStorage = key => {
```

vipulnsward (approved):
Some name changes, please change and merge on green.

PR #1076, opened 2023-02-23
Updated Authenticable concern to support auth email and token 

prasanthchaduvula:
Closes

Why

What
- Updated Authenticable concern to support auth email and token from any client-side app like postman instead of bearer token.
- Fixed failing specs related to the removal of bearer tokens as we changed to auth token.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -83,7 +83,7 @@ RSpec.configure do |config|
     end
   end
 
-  def headers(auth_user, options = {})
+  def auth_headers(auth_user, options = {})
     {
       "X-Auth-Token" => auth_user[:token],
       "X-Auth-Email" => auth_user[:email]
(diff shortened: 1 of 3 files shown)


vipulnsward:
This looks like a big change to me, @apoorv1316 please do a smoke test once locally, if all is good and CI is green we can merge.

apoorv1316:
@prasanthchaduvula @vipulnsward LGTM.

spec/rails_helper.rb
```diff
@@ -82,4 +82,11 @@
       Capybara.server_port = 3000
     end
   end
+
+  def headers(auth_user, options = {})
```
vipulnsward, line 86: Lets rename to auth_headers or something more descriptive?

vipulnsward (approved):
One comment on method name

PR #1079, opened 2023-02-24
Removed registerIntercepts, setAuthHeaders from the components and axios.ts file

prasanthchaduvula:
Closes

Why

What

- Removed registerIntercepts and setAuthHeaders from all the components as they are handled by request interceptors of `api.ts`
- Deleted `axios.ts` file as we are using `api.ts` to handle requests


vipulnsward (approved):
lgtm

PR #1082, opened 2023-02-24
Add, remove auth token & email on Login and Logout respectively

prasanthchaduvula:
Closes

Why

What

- Added AuthProvider to `App` component
- Login the user on the frontend so that in future it will be helpful when we include `Authenticable` concern at controller level
- Remove auth token and email on logout

### DIFF
--- a/app/javascript/src/components/Main.tsx
+++ b/app/javascript/src/components/Main.tsx
@@ -1,10 +1,32 @@
-import React from "react";
+import React, { useEffect } from "react";
 
 import { Roles } from "constants/index";
+import { useAuthDispatch } from "context/auth";
 
 import Dashboard from "./Dashboard";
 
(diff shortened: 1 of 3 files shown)


app/javascript/src/components/Main.tsx
```diff
@@ -1,10 +1,32 @@
...
+const Main = (props: Iprops) => {
+  const authDispatch = useAuthDispatch();
+  //@ts-expect-error for authDispatch initial values
+  const { token, email } = props.user;
+
+  useEffect(() => {
+    handleAuthLogin();
+  }, [email, token]);
+
+  const handleAuthLogin = async () => {
```
prasanthchaduvula, line 17: `handleAuthLogin` will be moved to the `SignIn `component once we integrate auth API.

vipulnsward (approved):

PR #1086, opened 2023-02-26
Add setup to run specs in parallel on local

vipulnsward:
**What:** 
- Add setup to run specs in parallel on local
- Add README instructions for the same

**Why:**
- For faster feedback loop on local instead of waiting on CI

**Before:**

Example output without Parallel enabled:
```ruby
Finished in 2 minutes 53.2 seconds (files took 1.79 seconds to load)
1175 examples, 0 failures
```

**After**

Example output with Parallel enabled (M1 Macbook Pro with 10 processes):
```ruby
1175 examples, 1 failure

Took 89 seconds (1:29)
Tests Failed
```

Improvements for me = Specs take 51% time instead of before 3 minutes.

**Todo:** 
- Add instructions to run only unit or system specs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -222,3 +222,6 @@ gem "strong_migrations"
 # Error tracking: https://docs.sentry.io/platforms/ruby/guides/rails/
 gem "sentry-rails"
 gem "sentry-ruby"
+
+# https://github.com/grosser/parallel_tests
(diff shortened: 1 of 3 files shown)


vipulnsward:
@prasanthchaduvula  / @shalapatil please test once locally tomorrow

PR #1096, opened 2023-03-01
Add Index prefix to ES to avoid name collisions in Index

### LINKED ISSUE #1095: Add ElasticSearchPrefix for Staging / Production Cluster
### Description
Set  ELASTIC_SEARCH_PREFIX: miru_production

In ConfigMap for Miru Production / Staging Accordingly

vipulnsward:
Fixes #1095 

What:
- Add Index prefix to ES to avoid name collisions in Index
- Add ENV as prefix 

Why:
- To avoid name collisions when sharing same ES

### DIFF
--- a/config/initializers/elasticsearch.rb
+++ b/config/initializers/elasticsearch.rb
@@ -2,3 +2,4 @@
 
 Searchkick.client = Elasticsearch::Client.new(url: ENV["ELASTICSEARCH_URL"])
 Searchkick.search_timeout = 300
+Searchkick.index_prefix = "Miru"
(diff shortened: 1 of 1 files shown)


config/initializers/elasticsearch.rb
```diff
@@ -2,3 +2,4 @@
 
 Searchkick.client = Elasticsearch::Client.new(url: ENV["ELASTICSEARCH_URL"])
 Searchkick.search_timeout = 300
+Searchkick.index_prefix = "Miru_#{Rails.env}_"
```
shalapatil, line 5: what's the value for Rails.env for staging?

PR #1110, opened 2023-03-06
System specs for clients

apoorv1316:
What
- System specs for clients


spec/system/clients/edit_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    login_as(user)
+  end
+
+  context "when editing a client" do
+    it "edit the client successfully" do
+      with_forgery_protection do
+        visit "/clients"
+
+        find(:xpath, '//tr[@class=" hoverIcon"]').hover.click()
+        find("[data-cy='three-dots']").click()
```
mohinid, line 22: Hi @apoorv1316 , "data-cy" tag was added for cypress specifically, should not use in this suite. Can you please check with other xpath. Can discuss more on slack.

apoorv1316: @mohinid Replaced it with an id selector.

spec/system/clients/delete_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  context "when deleting a client" do
+    it "delete the client successfully" do
+      with_forgery_protection do
+        visit "/clients"
+
+        expect(page).to have_content(client.name)
+
+        find(:xpath, '//tr[@class=" hoverIcon"]').hover.click()
```
mayankagnihotri7, line 23: Make use of something like `find(:css, ". hoverIcon", visible: false).hover.click`.

apoorv1316: Using visible with a boolean is not recommended. But I have replaced the XPath with CSS. https://docs.rubocop.org/rubocop-capybara/cops_capybara.html#capybaravisibilitymatcher

prasanthchaduvula:
@apoorv1316 I fixed the outstanding amount calculation on spec. You can continue your work. Removed the fronted cashFormatting part so that spec will pass and in future we can integrate it.

apoorv1316:
@prasanthchaduvula Please review

prasanthchaduvula:
@apoorv1316 Please fix the failing system specs due` login_as` method

apoorv1316:
@prasanthchaduvula I have fixed the sign_in method but the specs are failing because ES is not working and one reset password spec which is written by you is failing. Please check.

vipulnsward (approved):

PR #1111, opened 2023-03-06
Bugfix- Time entry report UI issue

onkar-saeloun:
What-
The PR includes time entry report UI issues Fixes on time entry report and filter page

Why- 
The font size of the responsive design was not matching with the design.
The view was not matching while running on a mobile device.

Notion- 
https://www.notion.so/saeloun/Mobile-responsive-page-Reports-UI-review-33440fabeb634c598c6183cad04f6e7d?pvs=4

screenshots-

Screenshot 2023-03-06 at 11 14 06 AM
Screenshot 2023-03-06 at 11 13 38 AM
Screenshot 2023-03-06 at 11 12 59 AM


app/javascript/src/common/ClickableCheckboxText.tsx
```diff
@@ -0,0 +1,102 @@
+import React from "react";
```
onkar-saeloun, line 1: Duplicated the file because as per the feedback whole row should be clickable.

PR #1122, opened 2023-03-10
Employees should not be able to access org. settings

deepti-kakade:
**Notion:**
https://www.notion.so/saeloun/6afc9500b2cc42af86be55f37894fb09?v=185bd45b9dc24dfcad9c57c07d32a7c0&p=1ba8f31d4c3149f79dcc5d2cfc29eb8b&pm=s

Description:

Currently, employees can access the org. settings page by changing the URL from the browser. Only admins and owners should be able to access the org. settings page.

- Updated company policy for index action. Only the admin and owner can access this page.

- Updated index specs.

Screenshot 2023-03-10 at 3 46 04 PM

PR #1133, opened 2023-03-14
Fixed workspace navigation for employee

prasanthchaduvula:
Closes 
 https://www.notion.so/saeloun/6afc9500b2cc42af86be55f37894fb09?v=b4528f26f4424af7beaf036262796cb3&p=74dbc8c31d2745f68668921e1f8007b0&pm=s
 
 ### Why
 On current development if an employee logs in companies controller index action throws an error ->. you are not authorised
Reason: In this PR https://github.com/saeloun/miru-web/pull/1122/files company policy we are only checking for the owner role and admin role and we are missing the employee role. But on the sidebar of the app, we are fetching companies which cause this issue

```
class CompanyPolicy < ApplicationPolicy
  attr_reader :error_message_key
  def index?
  end
end
```
On the `userActions` inside `app/javascript/src/components/Navbar/UserActions.tsx `we are using `companiesApi` index action for all the users

### What
- on `userActions` there is no need of fetching all the company details to display the current user details.
- With the help of the workspaces we can set the current user info.

### DIFF
--- a/app/javascript/src/components/Navbar/UserActions.tsx
+++ b/app/javascript/src/components/Navbar/UserActions.tsx
@@ -7,10 +7,10 @@ import { NavLink } from "react-router-dom";
 import { Avatar, Tooltip } from "StyledComponents";
 
 import authenticationApi from "apis/authentication";
-import companiesApi from "apis/companies";
 import WorkspaceApi from "apis/workspaces";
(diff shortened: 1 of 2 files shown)


onkar-saeloun (approved):
LGTM

PR #1141, opened 2023-03-15
Fix: Export to CSV/PDF functionality is not working on time entry report

mayankagnihotri7:
## Notion Card
[Notion Card](https://www.notion.so/saeloun/Export-to-CSV-PDF-functionality-is-not-working-on-time-entry-report-on-Chrome-Browser-f05e60b740814745a572810cc04ffb68)

## Description
- Fixed the bug where the export to CSV/PDF functionality was not working on the time entry report page.

## Preview

https://user-images.githubusercontent.com/57438322/225282589-3b051fca-a544-4695-89c0-544021a5f575.mp4


sanjibroy360 (changes requested):
You don't need to use `useRef` here.
Replace the `onClick` with the `onMouseDown` event.

``` js
 onMouseDown={() => {
     setShowExportOptions(false);
      handleDownload("csv");
  }}
  
  onMouseDown={() => {
     setShowExportOptions(false);
      handleDownload("pdf");
 }}

```
As `onMouseDown` fires immediately when the user clicks the button, while `onClick` fires after the button are released. By using `onMouseDown`, we can prevent the `onBlur` event from being fired before the click event has a chance to complete.

app/javascript/src/components/Reports/Header/index.tsx
```diff
@@ -40,6 +40,20 @@ const Header = ({
...
+    if (
+      menuRef.current &&
+      showExportOptions &&
+      !menuRef.current.contains(e.target)
+    ) {
+      setShowExportOptions(false);
+    }
+  };
+
+  document.addEventListener("mousedown", closeExportOptions);
```
onkar-saeloun, line 55: Mayank, can't we avoid using the document function here?
Instead, can we add the event listener to the div? 

e.g. `<div onMouseDown={closeExportOptions }></div>`

mayankagnihotri7:
@sanjibroy360 @onkar-saeloun I've made the changes. Please review.

sanjibroy360:
LGTM

sanjibroy360 (approved):

PR #1146, opened 2023-03-16
Fix: Incorrect client logo is displayed when the time entry report is grouped by team members

mayankagnihotri7:
## Notion Card
[Notion Card](https://www.notion.so/saeloun/Incorrect-client-logo-is-displayed-when-the-time-entry-report-is-grouped-by-team-members-654c9936eb6044d69705ec726f8b5f35)

## Description
- Fixed the bug where the incorrect client logo is displayed when the time entry report is grouped by team members and week.
- Fixed some warnings on the browser console whenever the filter was opened.

## Preview

https://user-images.githubusercontent.com/57438322/225597943-b4592657-394e-45ac-9f0a-3efb06a08b23.mp4

### DIFF
--- a/app/javascript/src/common/CustomRadio.tsx
+++ b/app/javascript/src/common/CustomRadio.tsx
@@ -40,12 +40,11 @@ const CustomRadioButton = ({
     <input
       checked={defaultCheck}
       className={classnames(DEFAULT_STYLE_INPUT, classNameInput)}
-      defaultChecked={defaultCheck}
       id={id}
       name={groupName}
       type="radio"
       value={value}
-      onClick={handleOnChange}
+      onChange={handleOnChange}
     />
     <label
       className={classnames("flex cursor-pointer items-center")}
--- a/app/javascript/src/components/Reports/Container/index.tsx
+++ b/app/javascript/src/components/Reports/Container/index.tsx
@@ -64,10 +64,10 @@ const Container = ({ selectedFilter }: ContainerProps) => {
     ) : null;
   };
 
-  const getEntryList = (entries, clientLogo: string) =>
+  const getEntryList = entries =>
     entries.map((timeEntry, index) => (
       <ReportRow
-        clientLogo={clientLogo}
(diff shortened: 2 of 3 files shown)


app/javascript/src/common/CustomRadio.tsx
```diff
@@ -40,12 +40,11 @@ const CustomRadioButton = ({
     <input
       checked={defaultCheck}
       className={classnames(DEFAULT_STYLE_INPUT, classNameInput)}
-      defaultChecked={defaultCheck}
```
Shruti-Apte, line 43: Why did you remove defaultChecked? 
We are using that prop in other files.

app/javascript/src/common/CustomRadio.tsx
```diff
@@ -40,12 +40,11 @@ const CustomRadioButton = ({
...
       id={id}
       name={groupName}
       type="radio"
       value={value}
-      onClick={handleOnChange}
```
Shruti-Apte, line 48: same for the onClick event, not sure but we must be using it somewhere as it is a common component. 
can you verify once if it's not affecting other components?

app/javascript/src/common/CustomRadio.tsx
mayankagnihotri7: I removed it because it does the same thing as checked. `defaultCheck` is a boolean value so only one of the `checked` or `defaultChecked` can be used.

app/javascript/src/common/CustomRadio.tsx
mayankagnihotri7: I changed `onClick` event to onChange because I removed `defaultChecked` attribute and `onClick` works when `checked` attribute isn't present and since we had both it was giving warnings in the console. When using checked attribute it needs `onChange`

supriya3105:
@Shruti-Apte Please resolve review comments if the responses are accepted. Approve accordingly.

Shruti-Apte (approved):

PR #1150, opened 2023-03-16
Removed unused devise routes

### LINKED ISSUE #1149: Remove unused devise routes
### Why
- As we moved auth routes to the client side and handled requests on internal API, there are a few routes on the backend for auth navigation on devise which are not needed any more

### What
- As incremental commit, navigate users to root_path who are intentionally visiting `users/sign_in` and `users_signup`
- Delete the password controller which is not needed any more as we have custom password controller inside internal API

prasanthchaduvula:
Closes #1149 

### Why
As we moved auth routes to the client side and handled requests on internal API, there are a few routes on the backend for auth navigation on devise which are not needed any more

### What
As an incremental commit, 
- Navigated users to root_path who are intentionally visiting users/sign_in and users_signup
- Fixed failing specs by navigating them to root path wherever there is session or registration path
- Delete the password controller which is not needed any more as we have a custom password controller inside the internal API. Removed devise password related routes, actions, views  and specs


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -8,7 +8,13 @@ def google_oauth2
...
     else
       flash[:error] = "There was a problem signing you in through Google. Please register or try signing in later."
-      redirect_to new_user_registration_url
+      redirect_to root_path
     end
   end
+
+  protected
+
+    def after_omniauth_failure_path_for(scope)
```
vipulnsward, line 17: Does this need a spec?

vipulnsward (approved):

app/controllers/users/omniauth_callbacks_controller.rb
prasanthchaduvula: No, devise supports it

prasanthchaduvula: Currently we are not using omniauth controller as google auth is not present

PR #1173, opened 2023-03-24
Invoice audit

shalapatil:
Notion: https://www.notion.so/saeloun/Backend-Admin-owner-user-should-be-able-to-view-the-change-log-for-an-invoice-844e71c756be4ffca24e192d65a7e039?pvs=4

What: Add API to get action trails for the invoice

Why: We want a new feature which will show the actions on the invoice 

- Created by user at ts
- Sent by user at ts to emails 
- Edited by user at ts
- Viewed by enduser at ts
- Download by user at ts
- Paid with stripe by enduser with email at ts 
- Paid by user at ts


shalapatil:
@vipulnsward I have considered basic add and update events for the trail. Please check all the requirements for the remaining events and let me know how should we go about custom columns.

db/migrate/20230321102347_create_versions.rb
```diff
@@ -0,0 +1,39 @@
...
+  # when serializing very large objects, `text` might not be big enough.
+  TEXT_BYTES = 1_073_741_823
+
+  def change
+    create_table :versions do |t|
+      t.string :item_type, null: false
+      t.bigint :item_id, null: false
+      t.string :event, null: false
+      t.string :whodunnit
+      t.text :object, limit: TEXT_BYTES
```
vipulnsward, line 18: Can we remove `object` here?  Lets check if everything else works fine in that case. We just want to track events and not object changes

shalapatil: @vipulnsward Yes, we can remove object column. 
There is one more column which tracks object_changes. We will need to parse that data in order to find out status changes like paid.

vipulnsward:
As discussed lets use https://github.com/ankane/ahoy instead?

PR #1189, opened 2023-03-29
Fix: Undefined method pages on time entry report

mayankagnihotri7:
## Notion Card
[Notion Card](https://www.notion.so/saeloun/Undefined-method-pages-on-time-entry-report-15f156822624426eb55b9822a24a60ed?pvs=4)

## Description
To fix the following issues -
- Showing of pagination option even when no pages are present.
- Previous and Next buttons enabled even when on the first and the last page.
- Page crashes when the user filters by using group by week.
- Pagination shown in between content for the mobile view.

## Preview

https://user-images.githubusercontent.com/57438322/228483995-0332277a-9ab6-4ae7-9e38-087441910f19.mp4


apoorv1316:
@mayankagnihotri7 There are 124 entries in my local and all of them are shown on the first and no pagination is shown. What's the max number of entries that we want to show on a page?
Screenshot 2023-03-29 at 5 16 25 PM

apoorv1316 (approved):

PR #1205, opened 2023-04-02
Upgrade Rails and puma

vipulnsward:
What:
- Upgrade Rails and puma

Why:
- Chore


vipulnsward:
@gowsik-ragunath please review and deploy/test on staging.

gowsik-ragunath (approved):
LGTM

PR #1208, opened 2023-04-03
Update packages causing warning with react 18

onkar-saeloun:
What-
PR includes updated packages as per react 18v

why-
The packages were not compatible with react version 18.

Note-
The only waring remaining is of react phone input which will resolve in separate PR because it might need some extra effort.


vipulnsward:
@onkar-saeloun specs are failing, please check

prasanthchaduvula (approved):

PR #1218, opened 2023-04-05
New address changes in client modal

onkar-saeloun:
What-
PR includes new address field changes on adding a new client and updating a client.

Why-
The backend implementation of new address fields was breaking the UI.

Loom-
https://www.loom.com/share/8f5f4e63870a42809c6fcfffd01911a0


app/javascript/src/components/Clients/Modals/ClientForm.tsx
```diff
@@ -23,24 +31,71 @@ interface FormValues {
...
+  /^((\+\d{1,3}(-| )?\(?\d\)?(-| )?\d{1,3})|(\(?\d{2,3}\)?))(-| )?(\d{3,4})(-| )?(\d{4})(( x| ext)\d{1,5}){0,1}$/;
+
 const clientSchema = Yup.object().shape({
   name: Yup.string().required("Name cannot be blank"),
   email: Yup.string()
     .email("Invalid email ID")
     .required("Email ID cannot be blank"),
-  phone: Yup.number().typeError("Invalid phone number"),
-  address: Yup.string().required("Address cannot be blank"),
+  phone: Yup.string()
```
prasanthchaduvula, line 51: Move the client validation schema to another file

app/javascript/src/components/Clients/Modals/ClientForm.tsx
```diff
@@ -23,24 +31,71 @@ interface FormValues {
...
+
+  return "";
+};
+
 const getInitialvalues = (client?: any) => ({
   name: client?.name || "",
   email: client?.email || "",
   phone: client?.phone || "",
-  address: client?.address || "",
+  address1: client?.address?.address_line_1 || "",
```
prasanthchaduvula, line 81: Move this initial values to the schema file
kind of utils

prasanthchaduvula (changes requested):

app/javascript/src/components/Clients/Modals/ClientForm.tsx
onkar-saeloun: Done

app/javascript/src/components/Clients/Modals/ClientForm.tsx
onkar-saeloun: Done

app/javascript/src/components/Clients/Modals/NewClient.tsx
```diff
@@ -22,7 +22,34 @@ const EditClient = ({
     formData.append("client[name]", values.name);
     formData.append("client[email]", values.email);
     formData.append("client[phone]", values.phone);
-    formData.append("client[address]", values.address);
+
```
prasanthchaduvula, line 25: @onkar-saeloun Please extract the common code of `EditClient` and `NewClient`components like form and submit methods to `ClientForm`

prasanthchaduvula (changes requested):

app/javascript/src/components/Clients/Modals/NewClient.tsx
onkar-saeloun: ok

app/javascript/src/components/Clients/Modals/EditClient.tsx
```diff
@@ -16,17 +17,9 @@ const EditClient = ({ setShowEditDialog, client }: IEditClient) => {
   const [clientLogoUrl, setClientLogoUrl] = useState<string>(client.logo);
```
prasanthchaduvula, line 17: If possible please move the logo related common codebase to client form from edit client and new client components

app/javascript/src/components/Clients/Modals/schema.ts
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable import/exports-last */
+import { Country } from "country-state-city";
```
prasanthchaduvula, line 2: Rename this to `formValidationSchema`

prasanthchaduvula (changes requested):

app/javascript/src/components/Clients/Modals/schema.ts
onkar-saeloun: Okay

app/javascript/src/components/Clients/Modals/utils.ts
```diff
@@ -0,0 +1,43 @@
+export const formatFormData = (
```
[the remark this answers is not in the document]
onkar-saeloun: same as above. If we need to reuse the same form anywhere else then we can make use of client form, keeping the form submission step in it will make it work for only in this component

app/javascript/src/components/Clients/Modals/schema.ts
onkar-saeloun: Done

app/javascript/src/components/Clients/Modals/utils.ts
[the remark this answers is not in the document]
prasanthchaduvula: We won't use the same form anywhere. Please move it to the client form so that it's easy to maintain.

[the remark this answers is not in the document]
onkar-saeloun: @prasanthchaduvula comment resolved please have a look

app/javascript/src/components/Clients/Modals/ClientForm.tsx
```diff
@@ -89,10 +142,44 @@ const ClientForm = ({
...
+      clientData,
+      clientLogo,
+      clientLogoUrl
+    );
+    if (formType === "new") {
+      try {
+        const res = await clientApi.create(formData);
+        setClientData([...clientData, { ...res.data, minutes: 0 }]);
+        setnewClient(false);
+        document.location.reload();
```
prasanthchaduvula, line 161: Fetching the updated list of clients from the API after creating a new client is generally a better approach than reloading the page using document.location.reload(). Reloading the page causes the entire page to refresh, which can be disruptive to the user experience and cause unnecessary load times.

On the other hand, fetching the updated list of clients from the API is faster and more efficient because it only retrieves the data that has changed. Additionally, fetching the updated data allows for a more seamless user experience, as the user does not need to wait for the page to reload to see the new client in the list.

**otherwise**

You can simply update the client list by using the response from create mclient or edit the client and appending it to the list state.

app/javascript/src/components/Clients/Modals/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  formData,
+  values,
+  isNewForm,
+  client,
+  clientLogo,
+  clientLogoUrl
+) => {
+  formData.append("client[name]", values.name);
```
prasanthchaduvula, line 9: Please make this change
https://github.com/saeloun/miru-web/pull/1218/files#r1162730052

prasanthchaduvula (changes requested):

app/javascript/src/components/Clients/Modals/utils.ts
onkar-saeloun: @prasanthchaduvula didn't get what's the issue? keeping it in different file is good thing right? Otherwise component will become complex

prasanthchaduvula: The formData is used in client form for both create and edit. It's better to move the logic to the same component. It's easy for the developers to read, maintain and debug.

prasanthchaduvula: Lets keep it. In future if we feel its hard to debug we will make the change

app/javascript/src/components/Clients/Modals/ClientForm.tsx
onkar-saeloun: Yah removed document.location.reload(). Managed the API with state change

onkar-saeloun:
@prasanthchaduvula Done with the change.

prasanthchaduvula (approved):

PR #1219, opened 2023-04-05
Fly production config

gowsik-ragunath:
What changed:

- Added fly config for miru production

Why:

- Migrated the production site from AWS to Fly.io


.node-version
```diff
@@ -1 +1 @@
-18.14.2
+18.14.2
```
vipulnsward, line 1: Please revert this

vipulnsward (approved):

PR #1224, opened 2023-04-06
Added fix for react phone input package

onkar-saeloun:
What- Added fix for React phone input package deprecation warning.

Why- React phone input version was not supporting React 18 and giving a deprecation warning while installing yarn.


onkar-saeloun:
@apoorv1316 can you please test this PR by visiting Team Member's edit page? It might be possible that the packages won't give any issue on my machine so better to test this.

apoorv1316 (approved):

PR #1235, opened 2023-04-10
mobile view project list page

Shruti-Apte:
### **Notion :**
https://www.notion.so/saeloun/Mobile-Responsive-Projects-List-page-bc3f85d80a2c48368d3862a89e095bb6?pvs=4
### **What :**
- Responsive project list page
### **Why :**
- Enhancement and requirement.
### **Preview :**
Screenshot 2023-04-10 at 8 29 18 PM

Screenshot 2023-04-11 at 11 46 28 AM

### **Todos** (for testing):
- Need to test with multiple clients and projects.

### DIFF
--- a/app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobile.tsx
+++ b/app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobile.tsx
@@ -1,0 +1,252 @@
+import React, { useState, useEffect } from "react";
+
+import { Formik, Form, FormikProps } from "formik";
+import Logger from "js-logger";
+import { XIcon, SearchIcon } from "miruIcons";
+import { Button, MobileMoreOptions } from "StyledComponents";
+
+import projectApi from "apis/projects";
+import CustomRadioButton from "common/CustomRadio";
+import { InputField, InputErrors } from "common/FormikFields";
+
+const AddEditProjectMobile = ({
+  editProjectData,
+  setEditProjectData,
+  setShowProjectModal,
+}) => {
+  const [client, setClient] = useState<object>({});
+  const [clientName, setClientName] = useState<string>("");
+  const [clientList, setClientList] = useState<any[]>([]);
+  const [showClientList, setShowClientList] = useState<boolean>(false);
+
+  const projectId =
+    (editProjectData && editProjectData["id"]) ||
+    Number(window.location.pathname.split("/").at(-1));
+
+  const isEdit = !!projectId;
+
+  interface FormValues {
+    client: string;
+    project: string;
+    isBillable: boolean;
+  }
+
+  const initialValues = {
+    client: editProjectData.clientName,
+    project: editProjectData.name,
+    isBillable: editProjectData.isBillable || editProjectData.is_billable,
+  };
+
+  const getClientList = async () => {
+    try {
+      const { data } = await projectApi.get();
+      setClientList(data.clients);
+    } catch (error) {
+      Logger.error(error);
+    }
+  };
+
+  const getProject = async () => {
+    if (!editProjectData.members) {
+      try {
+        const { data } = await projectApi.show(projectId);
+        setEditProjectData(data.project_details);
+      } catch (error) {
+        Logger.error(error);
+      }
+    }
+  };
+
+  const createProject = async values => {
+    await projectApi.create({
+      project: {
+        client_id: client["id"],
+        name: values.project,
+        billable: values.isBillable,
+      },
+    });
+    setEditProjectData("");
+    window.location.reload();
+    setShowProjectModal(false);
+  };
+
+  const editProject = async values => {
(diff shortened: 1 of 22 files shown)


app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobile.tsx
```diff
@@ -0,0 +1,285 @@
...
+import { useDebounce } from "helpers";
+import Logger from "js-logger";
+import { XIcon, SearchIcon } from "miruIcons";
+import { Button, MobileMoreOptions } from "StyledComponents";
+
+import projectApi from "apis/projects";
+import CustomRadioButton from "common/CustomRadio";
+import { InputField, InputErrors } from "common/FormikFields";
+
+const AddEditProjectMobile = ({
```
prasanthchaduvula, line 13: Can we rename it to `ProjectMobileForm`

app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobile.tsx
```diff
@@ -0,0 +1,285 @@
...
+        ? editProjectData.isBillable
+        : true,
+  };
+
+  const getClientList = async () => {
+    try {
+      const { data } = await projectApi.get();
+      setClientList(data.clients);
+    } catch (error) {
+      Logger.error(error);
```
prasanthchaduvula, line 51: Don't use Logger

app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobile.tsx
```diff
@@ -0,0 +1,285 @@
...
+    }
+  };
+
+  const getProject = async () => {
+    if (!editProjectData.members) {
+      try {
+        const { data } = await projectApi.show(projectId);
+        setEditProjectData(data.project_details);
+      } catch (error) {
+        Logger.error(error);
```
prasanthchaduvula, line 61: Ditto
Don't use Logger

app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobile.tsx
```diff
@@ -0,0 +1,285 @@
...
+  const createProject = async values => {
+    await projectApi.create({
+      project: {
+        client_id: client["id"],
+        name: values.project,
+        billable: values.isBillable,
+      },
+    });
+    setEditProjectData("");
+    window.location.reload();
```
prasanthchaduvula, line 75: Fetching the updated list of projects from the API after creating a new project is generally a better approach than reloading the page using window.location.reload(). Reloading the page causes the entire page to refresh, which can be disruptive to the user experience and cause unnecessary load times.

On the other hand, fetching the updated list of projects from the API is faster and more efficient because it only retrieves the data that has changed. Additionally, fetching the updated data allows for a more seamless user experience, as the user does not need to wait for the page to reload to see the new project in the list.

You can simply updated the project list by using response from create project or by fetching the project list again

prasanthchaduvula (changes requested):

app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobile.tsx
Shruti-Apte: `ProjectMobileForm ` doesn’t define what this form does, It is confusing. 
We have multiple pages like the list page, the details page(future scope).
How about `AddEditProjectMobileForm`?

app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobileForm.tsx
```diff
@@ -0,0 +1,285 @@
+import React, { useState, useEffect } from "react";
```
prasanthchaduvula, line 1: Please rename the component to `ProjectForm`

app/javascript/src/components/Projects/List/Mobile/AddEditProjectMobileForm.tsx
```diff
@@ -0,0 +1,285 @@
...
+        ? editProjectData.isBillable
+        : true,
+  };
+
+  const getClientList = async () => {
+    try {
+      const { data } = await projectApi.get();
+      setClientList(data.clients);
+    } catch {
+      // error logic already handled in api.ts
```
prasanthchaduvula, line 51: This is wrong. If there is nothing to handle in a catch block then using a try-catch block doesn't make any sense. I think already api.ts already handles errors. Please verify and move away from try-catch as they are empty.

prasanthchaduvula (changes requested):

prasanthchaduvula:
As provided review comments earlier please resolve all the comments and reassign.

prasanthchaduvula (approved):

PR #1244, opened 2023-04-14
added personal details new design

onkar-saeloun:
What-
Added changes for personal details UI

Why-
The new changes was not there.

Note- Mobile responsive design is not there just need to add it once web design is merged.

Screenshot-
Screenshot 2023-04-14 at 12 06 03 PM
Screenshot 2023-04-14 at 12 05 46 PM

Notion-
https://www.notion.so/saeloun/Replace-the-new-designs-of-personal-details-page-when-user-clicks-on-settings-faea04b6bdb04eb08efd9340fcf79af8?pvs=4


app/javascript/src/components/Profile/UserDetail/Edit/validationSchema.ts
```diff
@@ -0,0 +1,46 @@
...
+    .when("is_email", {
+      is: true,
+      then: Yup.string().email("Please enter valid email"),
+    }),
+  changePassword: Yup.boolean(),
+  password: Yup.string().when("changePassword", {
+    is: true,
+    then: Yup.string()
+      .required("Please enter password")
+      .matches(/^\S.*\S$/, "Password can not start or end with a blank space")
```
apoorv1316, line 25: @prasanthchaduvula Added the validations here. earlier it allowed us to edit passwords and we only checked if our password matched. Now we have all the password validations that we used on the signup flow.

PR #1260, opened 2023-04-21
Create address for new users

apoorv1316:
What
- Create an Address for new signed-up/invited users.

Why
- We were not creating addresses for newly signed-up users and instead trying to update the address and it was throwing error


prasanthchaduvula (approved):

prasanthchaduvula:
@apoorv1316 Merge it once checks passes

PR #1326, opened 2023-05-17
Add safe operator in migrations

apoorv1316:


gowsik-ragunath (approved):
LGTM!

In the future, we can write operators like this in data migration. As we already have the data migration gem we should write update operations in data migration.

PR #1331, opened 2023-05-22
Update dockerignore

apoorv1316:
What
- Updated `dockeignore`  to include `public/packs` and `public/packs/test`

Why
- To reduce the size of build images


gowsik-ragunath (approved):
LGTM!

PR #1333, opened 2023-05-22
Hide org settings link for non admins mobile

apoorv1316:
What
- Hide org. settings link for non-admin users on mobile view

### Before

Screenshot 2023-05-22 at 4 11 46 PM

### After-:

Employees
- 
Screenshot 2023-05-22 at 4 09 54 PM

Admins
Screenshot 2023-05-22 at 4 10 29 PM


app/javascript/src/components/Profile/Layout.tsx
```diff
@@ -50,12 +50,12 @@ const Layout = ({ isAdminUser, user, company }) => {
               />
             </div>
             <div className="col-span-9">
-              <RouteConfig />
+              <RouteConfig isAdmin={isAdminUser} />
```
prasanthchaduvula, line 53: Instead of passing `isAdminUser` as prop to RouteConfig, we can directly import it from `userContext` in` RouteConfig`

app/javascript/src/components/Profile/Layout/MobileNav.tsx
```diff
@@ -71,7 +86,7 @@ const MobileNav = () => {
       <UserDetails />
       <TeamUrl
         currentWorkspaceName={currentWorkspace?.name || ""}
-        urlList={urlList}
+        urlList={isAdmin ? AdminUrlList : EmployeeUrlList}
```
prasanthchaduvula, line 89: Please fix the navigation on back button when user visits the profile settings page or org settings page and clicks on back arrow it navigates to time tracking instead of settings

app/javascript/src/components/Profile/RouteConfig.tsx
```diff
@@ -26,7 +26,7 @@ const RouteConfig = () => (
       <Route element={<LeaveBalance />} path="leave-balance" />
       <Route element={<OrgDetails />} path="organization-details" />
       <Route element={<UserDetailsEdit />} path="change" />
-      <Route element={<MobileNav />} path="option" />
+      <Route element={<MobileNav isAdmin={isAdmin} />} path="option" />
```
prasanthchaduvula, line 29: On the desktop, employee users can still navigate to the Org URL directly by using the browser URL which sends the API request to the backend. We are showing a toastr message saying that you are unauthenticated to view this page, which is a backend response. Please handle this routing on the frontend so that when an employee changes the browser URL to the org URL instead of sending the API request to the backend and showing half of the org setting page use react routing and show them 404 page

prasanthchaduvula (changes requested):

app/javascript/src/components/Profile/Layout.tsx
apoorv1316: Done

app/javascript/src/components/Profile/Layout/MobileNav.tsx
apoorv1316: Now redirecting user to 'profile/edit/option' when clicked on back arrrow

prasanthchaduvula (approved):

PR #1340, opened 2023-05-24
Rake task for using prod entries for testing

apoorv1316:
What
- Added Rake task for using prod entries for testing in local and staging environment


apoorv1316:
@prasanthchaduvula Please review it.

PR #1349, opened 2023-05-30
Mark Invoice as waived

gowsik-ragunath:
Notion: 

https://www.notion.so/As-an-admin-or-owner-user-I-should-be-able-to-waive-off-an-invoice-so-that-I-can-get-the-exact-outs-69abd57487b645ddbec8943b18a43077

What changed:

- Added controller and policy
- Added waived-off option on the invoice page
- In invoice mail and invoice page the amount will be striked when the invoice is waived

Demo:

https://youtu.be/VYK8QTq85Ls


prasanthchaduvula (approved):

apoorv1316:
@gowsik-ragunath Please add system specs for it.

PR #1352, opened 2023-05-31
Refactor client forms

prasanthchaduvula:
### Why
The current client forms and details components code base is not structured properly. codebase is not readable due to which adding new features or fixing any bug is always prone to bugs.

### What
- Split the existing client form components into multiple components based on the functionality.
- Refactored client details components.
- Extracted the common codebase into reusable components.
- Removed reload on updating clients.


app/javascript/src/components/Clients/ClientForm/UploadLogo.tsx
```diff
@@ -0,0 +1,144 @@
...
+    const fileExtensions = file.type.split("/")[1];
+    const validFileByteSize = "30000";
+    const fileSize = file.size;
+
+    return {
+      fileExtension: validExtensions.includes(fileExtensions),
+      fileSizeValid: fileSize <= validFileByteSize,
+    };
+  };
+
```
apoorv1316, line 47: @prasanthchaduvula We are uploading images at multiple places in the app. Can we move this function to utils?

app/javascript/src/components/Clients/ClientForm/index.tsx
```diff
@@ -152,85 +96,23 @@ const ClientForm = ({
...
-        .catch(e => {
-          setApiError(e.message);
-        });
+      try {
+        await clientApi.update(client.id, formData);
+        setShowEditDialog(false);
+        fetchDetails();
+      } catch (error) {
+        setApiError(error.message);
+        setSubmitting(false);
```
apoorv1316, line 108: Do we really need this?

app/javascript/src/components/Clients/ClientForm/UploadLogo.tsx
prasanthchaduvula: As of now, I made a common upload component for both new clients and edit clients.
Once we refactor other forms, we can extract them based on the common functionality.

app/javascript/src/components/Clients/Details/TotalHoursChart.tsx
```diff
@@ -0,0 +1,66 @@
...
+        currencySymb + cashFormatter(overdueOutstandingAmount?.overdue_amount),
+    },
+    {
+      title: "OUTSTANDING",
+      amount:
+        currencySymb +
+        cashFormatter(overdueOutstandingAmount?.outstanding_amount),
+    },
+  ];
+
```
apoorv1316, line 29: While refactoring projects I moved this to projects/constants. Can you import and use it here or move that code to common constants for app and then use it here.

apoorv1316 (approved):

app/javascript/src/components/Clients/Details/TotalHoursChart.tsx
prasanthchaduvula: Let me check the styling of these components

PR #1366, opened 2023-06-09
Disabled invoice view and restirct no. of recipient emails

gowsik-ragunath:
This changes are build on main branch code and not on develop branch code.

The reason is as the develop branch is not tested yet I'm using main branch code to implement this fix.

I will test this change in staging and deploy it to production.

Screenshot:

Screenshot from 2023-06-09 15-56-52


prasanthchaduvula:
Need to refactor the spam-related changes from the user model, especially the spam errors

gowsik-ragunath:
Sure will do that after deploying the fix to production

PR #1369, opened 2023-06-12
Add client role

apoorv1316:


app/controllers/internal_api/v1/clients_controller.rb
```diff
@@ -35,8 +41,13 @@ def show
 
   def update
     authorize client
-
-    if client.update!(update_client_params)
+    ActiveRecord::Base.transaction do
+      client.update!(update_client_params)
+      if update_client_params[:email]
+        user = User.find_by!(email: update_client_params[:email])
+        client_member = ClientMember.find_by!(client:)
```
prasanthchaduvula, line 48: Find the existing `client_member ` with `client` and `user`. otherwise, it will fail if we use it from the postman or any other service if we are only finding client member with only client

app/controllers/internal_api/v1/team_controller.rb
```diff
@@ -11,7 +11,6 @@ def index
       .ransack(first_name_or_last_name_or_recipient_email_cont: params.dig(:q, :first_name_or_last_name_or_email_cont))
     teams = query.result(distinct: true)
     invitations = invitations_query.result(distinct: true)
-
     render :index, locals: TeamPresenter.new(teams, invitations, current_user, current_company).index_data, status: :ok
```
prasanthchaduvula, line 14: I think this file is unnecessarily pushed without any changes

prasanthchaduvula (changes requested):

app/controllers/internal_api/v1/clients_controller.rb
apoorv1316: done

prasanthchaduvula (approved):

PR #1371, opened 2023-06-13
View stripe event log invoice

gowsik-ragunath:
What changed:

- Payment record will be created based on the stripe event.
- Checkout expired and payment failed event will be captured.
- Payment name will store the name on the card.


app/services/invoice_payment/stripe_checkout_failed.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_accessor :invoice
+
+  def initialize(event)
+    @event = event
+    @data_object = event.data.object
+    @invoice = nil
+  end
+
+  def process
+    @invoice = Invoice.find(event.data.object.metadata.invoice_id)
```
prasanthchaduvula, line 14: what if the invoice is not found? Please handle it

config/routes.rb
```diff
@@ -22,13 +22,13 @@ def draw(routes_name)
...
 
   Sidekiq::Web.use Rack::Auth::Basic do |username, password|
-    ActiveSupport::SecurityUtils.secure_compare(
-      ::Digest::SHA256.hexdigest(username),
-      ::Digest::SHA256.hexdigest(ENV["SIDEKIQ_USERNAME"])) &
       ActiveSupport::SecurityUtils.secure_compare(
-        ::Digest::SHA256.hexdigest(password),
-        ::Digest::SHA256.hexdigest(ENV["SIDEKIQ_PASSWORD"]))
-  end
+        ::Digest::SHA256.hexdigest(username),
```
prasanthchaduvula, line 26: No need to push this file.

prasanthchaduvula (changes requested):

prasanthchaduvula:
@gowsik-ragunath Without frontend changes, will it break the existing flow of the app

gowsik-ragunath:
> Without frontend changes, will it break the existing flow of the app

Yes, we just add logic to handle new events, it wont affect any existing flow

prasanthchaduvula (approved):

PR #1375, opened 2023-06-15
Administrator dashboard

gowsik-ragunath:
What changed:

- Added admin dashboard
- Only super admin user can access this page

Demo: https://youtu.be/3VH_xTc8Mno


prasanthchaduvula:
@gowsik-ragunath Please resolve conflicts
We should use react for the views instead of rails views

gowsik-ragunath:
@prasanthchaduvula this is a view page generated by gem which is separate from our normal app view and used only by super admin, we don't need to develop these pages in react.

prasanthchaduvula (approved):

prasanthchaduvula:
@apoorv1316 Please test this PR before merging

PR #1380, opened 2023-06-20
Re-enabled sign up

gowsik-ragunath:
What changed:

Re-enabled user sign up

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -139,6 +139,9 @@ ActiveRecord::Schema[7.0].define(version: 2023_06_12_105554) do
     t.datetime "updated_at", null: false
   end
 
+  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
+  end
+
   create_table "devices", force: :cascade do |t|
     t.bigint "user_id", null: false
     t.bigint "company_id", null: false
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -139,6 +139,9 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
prasanthchaduvula, line 142: Please check why this table got added to schema

lib/authentication/google.rb
```diff
@@ -11,29 +11,29 @@ def initialize(user_data)
...
+        else
+          user_first_name = first_name || email.split("@").first
+          user = User.new(
+            email:,
+            first_name: user_first_name,
+            last_name:,
+            password: Devise.friendly_token[0, 20]
+          )
+          user.skip_confirmation!
+          user.save
```
prasanthchaduvula, line 33: use bang here `user.save!`

prasanthchaduvula (changes requested):

db/schema.rb
gowsik-ragunath: This is a table created by data migration gem, as it is a part of our rails app it is added in schema.

[ref.](https://github.com/ilyakatz/data-migrate#installation)

prasanthchaduvula (approved):

apoorv1316 (approved):

PR #1382, opened 2023-06-20
Fix no invoices errors on client

prasanthchaduvula:
### What
-  Send an error response if no client is associated with the user with role client.
- Don't show search input if no invoices present
- Send an error response if now invoice is found with an external key
- Render `<NoInvoices />`  component if the invoice is not present with the external key
- Removed extra wrapper around invoice view which adds unnecessary margin

### DIFF
--- a/app/controllers/internal_api/v1/clients/invoices_controller.rb
+++ b/app/controllers/internal_api/v1/clients/invoices_controller.rb
@@ -13,6 +13,11 @@ class InternalApi::V1::Clients::InvoicesController < InternalApi::V1::Applicatio
   end
 
   def client
-    @_client ||= ClientMember.find_by!(user: current_user).client
+    client_member = ClientMember.find_by(user: current_user)
+    unless client_member && client_member.client
+      render json: { error: "No clients associated" }, status: :unprocessable_entity
+    else
+      @_client ||= client_member.client
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/internal_api/v1/clients/invoices_controller.rb
```diff
@@ -13,6 +13,11 @@ def index
   end
 
   def client
-    @_client ||= ClientMember.find_by!(user: current_user).client
+    client_member = ClientMember.find_by(user: current_user)
+    unless client_member && client_member.client
```
mayankagnihotri7, line 17: Can we use `if client_member && client_member.client` which checks for the true condition as opposed to the falsy condition being checked here?

app/javascript/src/components/ClientInvoices/Header.tsx
```diff
@@ -92,79 +92,82 @@ const Header = ({ params, setParams }) => {
...
-            />
-            <button className="absolute inset-y-0 right-3 flex cursor-pointer items-center">
-              {searchQuery ? (
-                <XIcon size={12} weight="bold" onClick={onSearchClear} />
-              ) : (
-                <SearchIcon
-                  className="text-miru-gray-1000"
-                  size={16}
-                  weight="bold"
+      {showSearch !== 0 && (
```
mayankagnihotri7, line 95: Instead of checking if it is not equal to zero, we can check like - `showSearch >= 1`

mayankagnihotri7 (changes requested):

app/controllers/internal_api/v1/clients/invoices_controller.rb
prasanthchaduvula: We need to send the response to client

mayankagnihotri7 (approved):

PR #1383, opened 2023-06-20
Cities not loading org setup

Shruti-Apte:
What: 
Cities not loading bug fix on org settings page

Why:
Earlier the list of city was empty.


apoorv1316 (changes requested):
@Shruti-Apte Cities are not loading when we don't change the default country. Please check
Screenshot 2023-06-21 at 10 35 32 AM

apoorv1316 (approved):

PR #1429, opened 2023-07-25
Refactored clients index action by moving away from ransack to searchkick

prasanthchaduvula:
### What
- We are using `Ransack` to search clients by name/email. With this PR, we will be using `Searchkick` to search clients by name/email. We are using `Searchkick ` in other places.

### DIFF
--- a/app/services/clients/index_service.rb
+++ b/app/services/clients/index_service.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module Clients
+  class IndexService < ApplicationService
+    attr_reader :params, :current_company
+
+    def initialize(current_company, params)
+      @current_company = current_company
+      @params = params
+    end
+
+    def process
+      {
+        client_details:,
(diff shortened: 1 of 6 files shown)


app/services/clients/index_service.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Clients
+  class IndexService < ApplicationService
+    attr_reader :params, :current_company
+
+    def initialize(current_company, params)
+      @current_company = current_company
+      @params = params
```
gowsik-ragunath, line 9: Instead of passing the entire params, can you pass only the `query` and `time_frame` values as argument to service? or pass a sanitized params as argument.

gowsik-ragunath (changes requested):
Added one minor comment

gowsik-ragunath (approved):
LGTM

apoorv1316 (approved):

PR #1438, opened 2023-07-31
Admin can add multiple recipients while sending invoice

prasanthchaduvula:
### What
- Supriya can add multiple recipients while sending invoice


gowsik-ragunath (approved):
LGTM

PR #1440, opened 2023-08-02
Autogenerate Invoice number

Shruti-Apte:
### **Notion :**
https://www.notion.so/saeloun/While-generating-a-new-invoice-the-invoice-number-should-be-autopulated-as-per-the-previous-invoice--05e98f0d18064bb5b47f9f4e92f8292d?pvs=4

### **What :**
- if the previous invoice number for a client is 003, then when the user selects the client name on a new invoice, then the invoice number field is getting be auto-populated with 004. 
- If last character of previous invoice number is alphabet then the invoice number will be blank.
- If no previous invoices are present for the selected client then the invoice number will be blank.
-
### **Why :**
- To ensure that the invoice number is incremented from the previous number and maintain consistency. 
### **Preview :**
https://www.loom.com/share/89f4ce5938f0498381582f247b797c41?sid=7a0ff982-fdb0-4258-a418-40a37311eb9a

### **Todos** (for testing):
- Tested according to requirement.

### DIFF
--- a/app/javascript/src/common/CustomInputText/index.tsx
+++ b/app/javascript/src/common/CustomInputText/index.tsx
@@ -8,7 +8,7 @@ const defaultInputBoxClassName =
   "form__input block w-full appearance-none bg-white p-4 text-sm lg:text-base h-12 border-miru-gray-1000";
 const defaultWrapperClassName = "outline relative h-12";
 const defaultLabelClassname =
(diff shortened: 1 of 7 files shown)


apoorv1316:
@Shruti-Apte Please fix the bug mentioned in the loom video.
https://www.loom.com/share/6baeef45bec04c92b73ef0cb2167b3f9

prasanthchaduvula (approved):

PR #1444, opened 2023-08-07
Hubspot form integration with background job

prasanthchaduvula:
### Notion :
https://www.notion.so/saeloun/Integration-with-Marketing-tool-Hubspot-4836e0626b8843378996ea943ccd1d13?pvs=4

### What :
- With this PR, any user who signups up on Miru either the owner who creates an organization or an invited user with any role will be added to the Hubspot form so that in future we can use those onboarded emails for communication purposes.
- Under the hood, it uses the background job to handle the Hubspot API integration asynchronously when a user transaction is completed 

### Why :
To keep track of user signups on Miru.

### Demo
https://youtu.be/gwj4frxs5eg

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -138,6 +138,9 @@ gem "ruby_audit", require: false
 gem "rubyzip"
 
 gem "ahoy_matey"
+
+gem "httparty"
+
 group :development, :test, :ci do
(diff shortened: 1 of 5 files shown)


gowsik-ragunath (approved):
LGTM

PR #1449, opened 2023-08-23
Update SessionsController's create action to send all the initial props to the Miru desktop app as login response

sanjibroy360:
## What
- Update SessionsController's create action to send all the initial props to the Miru desktop app as login response and also added specs for this change.
- It'll send the initial props as a response only when it has the `?app=miru_desktop` param otherwise it'll send the regular login response.

## Why
- To get the avatar_url and other initial props on the Miru desktop app as well.


prasanthchaduvula (commented):
@sanjibroy360 Let me test this out and will get back to you

prasanthchaduvula:
@sanjibroy360 I went through the changes and will try to push a change to avoid the skipping of skip_before_action :verify_authenticity_token, only: :create

prasanthchaduvula (approved):

PR #1477, opened 2023-09-21
Admin pto module

prasanthchaduvula:
#### Notion
https://www.notion.so/6afc9500b2cc42af86be55f37894fb09?v=b4528f26f4424af7beaf036262796cb3&p=2046f632b29c4be9b1d63ce66c935523&pm=s

#### What
- Admin or owners users should be add company level leaves settings i.e. can add, update and remove leaves based on the year

#### Demo
https://www.youtube.com/watch?v=Pb9DR4noxQ0


app/controllers/internal_api/v1/leave_types_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  private
+
+    def leave_type_params
+      params.require(:leave_type).permit(permitted_attributes)
+    end
+
+    def leave
+      @_leave ||= current_company.leaves.find_by!(id: params[:id])
```
gowsik-ragunath, line 26:
```suggestion
      @_leave ||= current_company.leaves.find(params[:id])
```

app/controllers/internal_api/v1/leave_types_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def leave_type
+      @_leave_type ||= LeaveType.find_by!(id: params[:id])
```
gowsik-ragunath, line 30:
```suggestion
      @_leave_type ||= LeaveType.find(params[:id])
```

app/controllers/internal_api/v1/leaves_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  private
+
+    def leave_params
+      params.require(:leave).permit(policy(:leave).permitted_attributes)
+    end
+
+    def leave
+      @_leave ||= Leave.find_by!(id: params[:id])
```
gowsik-ragunath, line 33:
```suggestion
      @_leave ||= Leave.find(params[:id])
```

app/policies/leave_policy.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class LeavePolicy < ApplicationPolicy
+  def index?
+    user_owner_role? || user_admin_role?
+  end
+
+  def create?
+    user_owner_role? || user_admin_role?
+  end
```
gowsik-ragunath, line 10: Can we move common conditions like this as a separate method?

```
def has_owner_or_admin_role?
end
```
```suggestion
  def index?
    has_owner_or_admin_role?
  end

  def create?
    has_owner_or_admin_role?
  end
```

app/policies/leave_type_policy.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class LeaveTypePolicy < ApplicationPolicy
+  def create?
+    user_owner_role? || user_admin_role?
```
gowsik-ragunath, line 5: Same here, you can even move `has_owner_or_admin_role?` method to application_policy.rb and use it when every it is needed.

app/policies/leave_with_leave_types_policy.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class LeaveWithLeaveTypesPolicy < ApplicationPolicy
+  def update?
+    user_owner_role? || user_admin_role?
```
gowsik-ragunath, line 5: Same here

app/controllers/internal_api/v1/leave_with_leave_types_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class InternalApi::V1::LeaveWithLeaveTypesController < InternalApi::V1::ApplicationController
+  before_action :leave, only: [:update]
+
+  def update
+    authorize current_user, policy_class: LeaveWithLeaveTypesPolicy
+    LeaveWithLeaveTypesService.new(leave, update_params).process
```
gowsik-ragunath, line 8: Instead of passing the whole update_params please pass only the value needed in the service, you can even create a separate hash with all the needed values.

gowsik-ragunath (changes requested):
Added a few minor comments

app/controllers/internal_api/v1/leave_with_leave_types_controller.rb
prasanthchaduvula: Currently, we are passing only the required values with the help of update_params by white-listing them

With the service, we are basically creating, updating and removing the leaves within the transaction

```ruby
def update_params
      params.require(:leave_with_leave_type).permit(
        add_leave_types: [:name, :color, :icon, :allocation_value,
                          :allocation_period, :allocation_frequency,
                          :carry_forward_days],
        updated_leave_types: [:id, :name, :color, :icon, :allocation_value,
                              :allocation_period, :allocation_frequency,
                              :carry_forward_days],
        removed_leave_type_ids: []
      )
    end
end

```

gowsik-ragunath: Even though `update_params` is a strong params the problem with pass the entire params is that in future, new keys can be introduced in this method and if those keys had any sensitive data, passing the entire `update_params` can lead to a security vulnerability.

It is always a better idea to break this params and pass only the needed values to services.

app/controllers/internal_api/v1/leave_types_controller.rb
prasanthchaduvula: Here we have to throw the error if leave is not found

gowsik-ragunath: Actually `find` does the same thing here, it is similar to `find_by!(id: )`, you can test it in rails console.

app/controllers/internal_api/v1/leave_with_leave_types_controller.rb
prasanthchaduvula: We are not going to add new keys as service deals with only these

```
add_leave_type
updated_leave_type
removed_leave_type_ids

```

gowsik-ragunath (approved):
LGTM!

PR #1537, opened 2023-10-09
Removed precompile command and added envs in dockerfile

gowsik-ragunath:
What changed:

- Removed the `rails assets:precompile` command from Dockerfile.local
- Added Bundler and Gem path ENVs

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -3,6 +3,10 @@ FROM ruby:3.2.2-slim AS base
 
 ARG NODE_VERSION=18.14.2
 
+ENV BUNDLE_PATH="/usr/local/bundle"
+ENV GEM_HOME="/usr/local/bundle/gems"
+ENV GEM_PATH="/usr/local/bundle/gems"
+
 RUN mkdir /app
 WORKDIR /app
 RUN mkdir -p tmp/pids
@@ -49,5 +53,3 @@ COPY --from=gems /app /app
 COPY --from=gems /usr/local/bundle /usr/local/bundle
 COPY --from=node_modules /app/node_modules /app/node_modules
 COPY . ./
-
-RUN RAILS_ENV=$RAILS_ENV rails assets:precompile


Dockerfile.local
```diff
@@ -3,6 +3,10 @@ FROM ruby:3.2.2-slim AS base
 
 ARG NODE_VERSION=18.14.2
 
+ENV BUNDLE_PATH="/usr/local/bundle"
+ENV GEM_HOME="/usr/local/bundle/gems"
+ENV GEM_PATH="/usr/local/bundle/gems"
```
akhilgkrishnan, line 8: Why do we need this change?

gowsik-ragunath: As the cached gems are stored in the `/usr/local/bundle/gems` path, we have to set this path in ENVs to avoid LoadError.

Screenshot 2023-10-09 at 6 43 06 PM



https://github.com/saeloun/miru-web/blob/bd50bce9815274f57d90a3aa222b1aae95472420/docker-compose.yml#L33

apoorv1316:
@gowsik-ragunath I was able to setup the app without any errors but the pages were not rendering. Then I tried running `bin/webpack-dev-server` command but it threw error 
`zsh: no such file or directory: bin/webpack`

Then I ran `bin/webpack` and the pages started loading after couple of refresh.

cc @akhilgkrishnan

akhilgkrishnan:
> @gowsik-ragunath I was able to setup the app without any errors but the pages were not rendering. Then I tried running `bin/webpack-dev-server` command but it threw error `zsh: no such file or directory: bin/webpack`
> 
> Then I ran `bin/webpack` and the pages started loading after couple of refresh.
> 
> cc @akhilgkrishnan

@apoorv1316 Why the webpacker is failing?

apoorv1316:
@akhilgkrishnan I tried reproducing it but it's working fine now. There were some errors in the browser console. Let's merge this PR?

akhilgkrishnan (approved):
LGTM

PR #1545, opened 2023-10-12
Dont allow duplicate client members with the same client and user

prasanthchaduvula:
### What
- We can't create a duplicate client member (client contact) with the same client and user if there is already a client member.


gowsik-ragunath (approved):
LGTM!

PR #1591, opened 2023-11-13
Add pagination styled component

prasanthchaduvula:
#### What
- Added the `Pagination` component to the `StyledComponents` component library.
- Added documentation of it to Miru docs.
- Used it in the `Invoices` , `teams`, `timesheet entry reports` and `client invoices`.
- Removed the common pagination which is not required any more.


apoorv1316 (changes requested):
@prasanthchaduvula On the invoices page if we change the value of per_page to 5 from URL then the page numbers are not displayed on the Pagination Component
Screenshot 2023-11-15 at 10 23 36 AM

apoorv1316 (approved):

PR #1593, opened 2023-11-15
Store bounced and compliant email address

gowsik-ragunath:
What changed:

- Added a controller which will store the bounced and compliant email addresses
- Re-enabled the weekly emails
- Added weekly email reminder to skip sending email to invalid email address that was bounced or compliant.

Why:

This is to prevent send mails to email address that raises the bounce and compliant rate


prasanthchaduvula:
@gowsik-ragunath In the file changes there is a controller and routing for `sns_subscriptions, only: [:create]` but nowhere do we trigger this endpoint in the file changes.

gowsik-ragunath:
@prasanthchaduvula This endpoint will be utilized by SNS event triggers. 

Whenever there is a hard-bounced email or a complaint report in AWS SES, a notification will be sent to this specific endpoint. In the controller, we will create a record with the email address included in the notification and associate it with the SesInvalidEmail model.

prasanthchaduvula (approved):

PR #1598, opened 2023-11-24
Fixed populating emails on send invoice

prasanthchaduvula:
#### What
- Updated the API response to include virtual verified emails and client contact emails for generate invoice.
- Made a common SendInvoice component and used it on the generate, edit, view and invoice list page.
- Removed the duplicate send invoice components.


apoorv1316 (commented):
@prasanthchaduvula Please work on mobile devices fixes too.

apoorv1316 (approved):

PR #1599, opened 2023-11-28
Invoice details UI fixed 

Shruti-Apte:
- Invoice details mobile view fixed 
- Buttons updated on where style prop was not passed


apoorv1316:
@Shruti-Apte Please add screenshot of before and after change.

PR #1609, opened 2023-12-20
Additional details for invitation email

apoorv1316:
Notion:

https://www.notion.so/saeloun/Backend-changes-Send-the-information-used-in-the-new-I-from-formatted-Invitation-email-39302fb836854979b7a200817de89167

### DIFF
--- a/app/javascript/src/components/Authentication/ResetPassword/index.tsx
+++ b/app/javascript/src/components/Authentication/ResetPassword/index.tsx
@@ -8,6 +8,7 @@ import authenticationApi from "apis/authentication";
 import { InputErrors, InputField } from "common/FormikFields";
 import MiruLogoWatermark from "common/MiruLogoWatermark";
 import { MIRU_APP_URL, Paths } from "constants/index";
+import { setToLocalStorage } from "utils/storage";
 
 import {
   resetPasswordFormInitialValues,
@@ -31,11 +32,8 @@ const ResetPassword = () => {
     };
     try {
       const res = await authenticationApi.resetPassword(payload);
-      if (res.status == 200) {
-        setTimeout(() => {
-          window.location.assign(`${window.location.origin}`);
-        }, 500);
-      }
+      setToLocalStorage("authToken", res.data.user.token);
+      setToLocalStorage("authEmail", res.data.user.email);
     } catch (error) {
       setFieldError("confirm_password", error.response.data.error);
     }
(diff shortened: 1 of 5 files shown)


app/models/invitation.rb
```diff
@@ -112,13 +112,26 @@ def recipient_email_not_changed
       end
     end
 
+    def users_not_with_client_role
+      users_with_client_role_ids = company.users.joins(:roles).where(roles: { name: "client" }).pluck(:id)
+      company.users.where.not(id: users_with_client_role_ids)
```
prasanthchaduvula, line 117: Do we have to check all the users without client including the discarded ones or only kept records
`company.users..keptwhere.not`

prasanthchaduvula (changes requested):

app/models/invitation.rb
apoorv1316: we want to display the count of employees in the org. Clients are included in employments but we don't want to count them.

app/javascript/src/components/Authentication/ResetPassword/index.tsx
```diff
@@ -31,11 +32,8 @@ const ResetPassword = () => {
     };
     try {
       const res = await authenticationApi.resetPassword(payload);
-      if (res.status == 200) {
-        setTimeout(() => {
-          window.location.assign(`${window.location.origin}`);
-        }, 500);
-      }
+      setToLocalStorage("authToken", res.data.user.token);
+      setToLocalStorage("authEmail", res.data.user.email);
```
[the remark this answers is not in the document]
apoorv1316: Added redirect

app/models/invitation.rb
prasanthchaduvula: I mean here we have to check kept users or all the users`company.users.kept`

apoorv1316: I found out the query was wrong. It was counting the invitations along with deleted ones. But we only want to show the employees. So the following query works and `kept`  is not required because we use `destroy` when deleting team members.
`company.employments.joins(user: :roles).where.not(roles: { name: "client" })`

apoorv1316:
@prasanthchaduvula Please review again

app/models/invitation.rb
prasanthchaduvula: But I think it should be discarded.

prasanthchaduvula (approved):

app/models/invitation.rb
prasanthchaduvula: Please check the team members controller destroy action as its discard the employee

PR #1612, opened 2023-12-27
Set expiry when invitation deleted

apoorv1316:
What
- Show the error message when the invitation is deleted
Screenshot 2024-01-03 at 7 40 20 AM

### DIFF
--- a/app/controllers/internal_api/v1/invitations_controller.rb
+++ b/app/controllers/internal_api/v1/invitations_controller.rb
@@ -19,6 +19,7 @@ class InternalApi::V1::InvitationsController < InternalApi::V1::ApplicationContr
   def destroy
     authorize @invitation
 
+    @invitation.mark_as_deleted
     @invitation.destroy!
     render json: { notice: I18n.t("invitation.delete.success.message") }, status: :ok
   end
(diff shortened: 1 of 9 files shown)


app/controllers/internal_api/v1/invitations_controller.rb
```diff
@@ -19,6 +19,7 @@ def update
   def destroy
     authorize @invitation
 
+    @invitation.mark_as_deleted
```
prasanthchaduvula, line 22: Remove this `mark_as_deleted`. Not required

app/models/invitation.rb
```diff
@@ -90,6 +90,11 @@ def set_token
     end
   end
 
+  def mark_as_deleted
```
prasanthchaduvula, line 93: Remove this not required

prasanthchaduvula (changes requested):

app/controllers/internal_api/v1/invitations_controller.rb
apoorv1316: done

app/models/invitation.rb
apoorv1316: done

prasanthchaduvula (approved):

PR #1655, opened 2024-01-18
Added leave management filters

prasanthchaduvula:
### What
- User can select the year and can view applied leaves and calculations based on the year.
- User can click on the leave types to view applied leaves with the same leave in that selected year along with the change in the calculation part.

### DIFF
--- a/app/javascript/src/components/LeaveManagement/index.tsx
+++ b/app/javascript/src/components/LeaveManagement/index.tsx
@@ -3,6 +3,7 @@ import React, { useEffect, useState } from "react";
 import { getYear } from "date-fns";
 
 import timeoffEntryApi from "apis/timeoffEntry";
+import Loader from "common/Loader/index";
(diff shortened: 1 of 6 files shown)


app/javascript/src/components/LeaveManagement/index.tsx
```diff
@@ -33,33 +42,70 @@ const LeaveManagement = () => {
...
+  const handlefilterTimeoffEntries = (
+    timeoffEntries,
+    totalTimeoffEntriesDuration
+  ) => {
+    if (selectedLeaveType) {
+      const sortedTimeoffEntries =
+        timeoffEntries.length &&
+        selectedLeaveType &&
+        timeoffEntries.filter(
+          timeoffEntry => timeoffEntry.leaveType.id == selectedLeaveType.id
```
JijoBose, line 63: let's use ===

JijoBose (approved):
LGTM

PR #1659, opened 2024-01-29
Discard leaves and holidays

prasanthchaduvula:
### What
- Discard leaves and leave types along with time off entries.
- Discard holidays and holiday info along with time off entries
- Updated the leave management calculations

### DIFF
--- a/app/models/holiday.rb
+++ b/app/models/holiday.rb
@@ -5,6 +5,7 @@
 # Table name: holidays
 #
 #  id                              :bigint           not null, primary key
+#  discarded_at                    :datetime
 #  enable_optional_holidays        :boolean          default(FALSE)
 #  holiday_types                   :string           default([]), is an Array
 #  no_of_allowed_optional_holidays :integer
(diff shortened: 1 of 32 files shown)


app/models/holiday.rb
```diff
@@ -16,18 +17,29 @@
...
+  include Discard::Model
+
   has_many :holiday_infos, dependent: :destroy
   belongs_to :company
 
   enum time_period_optional_holidays: { per_quarter: 0, per_year: 1, per_month: 2, per_week: 3 }
 
   validates :year, presence: true, uniqueness: { scope: :company_id }
   validates :year, numericality: { only_integer: true, greater_than_or_equal_to: 1900, less_than_or_equal_to: 2099 }
+
```
nisusam, line 37: Add default_scope

```RUBY
default_scope -> { kept }
```

nisusam (approved):
LGTM!

Add `default_scope -> { kept }` in respective models.

app/models/holiday.rb
prasanthchaduvula: @vipulnsward Can we add default scopes?

vipulnsward: We avoid default scope as much as possible.

PR #1668, opened 2024-02-23
Added week view with mark time off

prasanthchaduvula:
#### What
- With this PR, we can see the week view with timesheet entries and timeoff entries.
- We can add and update new rows and timesheet entries for the week with selected clients and projects.
- We can apply leaves by clicking mark time off and view. edit and delete all the leaves applied on the week.

#### Demo
https://youtu.be/xWXnOSL_5Ow?feature=shared


nisusam (approved):

apoorv1316:
@prasanthchaduvula After adding a entry in week view and then if i switch to day/month view I have to refresh the page to see the added entry. Please fix

Issue #1595, opened 2023-11-19
Can't add client - not sure what field is missing


anujdeshpande:
### Description
Unable to add a client to my Miru account

### Environment

Windows, Chrome (Brave)

### What is the expected behaviour?

Know which fields are mandatory or are malformed.

### What is the current behaviour?

Unable to add client, and can't figure out why.

### How to reproduce the issue?

1. Add contents as above
2. City dropdown is blank
3. Add zipcode, and press Save Changes. Save Changes is greyed out.

### Screenshots or Screencast

![image](https://github.com/saeloun/miru-web/assets/969376/653b39a2-deca-4ec6-a498-14f363cf65c0)
![image](https://github.com/saeloun/miru-web/assets/969376/d2b0363e-d766-4d1b-82c9-8426c0ad901a)

### If the bug is confirmed, would you be willing to submit a PR?

Sure

apoorv1316:
Hi @anujdeshpande, It seems to be an issue with the package we are using. It is not loading the list of cities for the United Kingdom country if you select any state except "England".  The temporary solution would be to select "England" as state and then you can get the list of cities and create a client. We will look into this and fix it soon. Thanks for opening the issue.

supriya3105:
@apoorv1316 Can you open an issue for it in Notion and add the details ?

anujdeshpande:
@supriya3105 I don't think this is fixed - is this only for self-installed instances?

supriya3105:
> @supriya3105 I don't think this is fixed - is this only for self-installed instances?

@apoorv1316 Can you respond to Anuj's query ?  
Also, if the issue is with the package , let's look at another package without this issue.

apoorv1316:
We have removed the dropdowns for state and city. It is now text fields. It will be released tomorrow

apoorv1316:
@anujdeshpande The issue is fixed now. Please refer to the loom video https://www.loom.com/share/2e4b5e5af67845539e5873e84414fd46

PR #1676, opened 2024-02-28
Remove validations for state, city and pin

adititipnis:


apoorv1316:
@adititipnis The [issue](https://github.com/saeloun/miru-web/issues/1595) which user reported still exists. How does removing validations fixes this issue? We can allow free text with the validations and we can either remove the dropdown or keep it while allowing the user to enter anything we want.

adititipnis:
> @adititipnis The [issue](https://github.com/saeloun/miru-web/issues/1595) which user reported still exists. How does removing validations fixes this issue? We can allow free text with the validations and we can either remove the dropdown or keep it while allowing the user to enter anything they want.

I am able to add a client with the details provided in the issue. The validations were blocking the client from being added since "city" was a required field which we cannot populate for Hampshire because of the limitations of the package. Such limitations exist for a few more states and cities so it makes sense to get rid of the validations to unblock similar scenarios. The only issue that remains is the 'undefined' response from the backend. This can be handled on the frontend but it would be more appropriate to receive an undefined type from the backend rather than a string as seen below.
image
image

vipulnsward:
- Do not keep drop downs for anything except country, don't complicate logic here to show textbox or dropdown
- Add validation that street address is mandatory
- Do not remove existing validations

PR #1680, opened 2024-02-28
Export options for accounts aging

apoorv1316:
What
- Added functionality to generate PDF for accounts aging
- Added functionality to generate CSV for accounts aging

Preview
https://www.loom.com/share/7bb851c20a7047c38523b0192bd064ec


app/services/reports/generate_csv.rb
```diff
@@ -0,0 +1,22 @@
...
+    @headers = headers
+  end
+
+  def process
+    CSV.generate do |csv|
+    csv << headers
+
+    data.each do |row|
+    csv << row
+  end
```
nisusam, line 19: Please reindent.

nisusam (changes requested):

app/services/reports/generate_pdf.rb
```diff
@@ -1,38 +1,38 @@
...
-    private
+  private
 
-      def generate_time_entries_pdf
-        Pdf::HtmlGenerator.new(
-          :reports,
-          locals: { report_data:, current_company: }
-        ).make
-      end
+    def generate_time_entries_pdf
```
nisusam, line 25: You can create common method for both the reports `def generate_pdf_reports(reports)`

nisusam (changes requested):

nisusam (approved):
@apoorv1316 The total amount is without currency in CSV Data. Please update the same in the new PR.

Rest LGTM!

PR #1685, opened 2024-03-03
Replace safe_email to with email in factories

apoorv1316:
Why
- safe_email is deprecated


nisusam (approved):
Fix failing spec. Rest LGTM!

PR #1698, opened 2024-03-08
Fix n+1 query in clients

apoorv1316:


app/services/clients/index_service.rb
```diff
@@ -23,9 +23,9 @@ def process
 
       def clients_list
         if query.present?
-          search_clients(search_term, where_clause)
+          search_clients(search_term, where_clause).includes(:logo_attachment)
```
nisusam, line 26: You can also assign this to `list` variable & return `list.includes(:logo_attachment)`

nisusam (approved):

PR #1699, opened 2024-03-12
Devices API 

apoorv1316:
What
- Add insurance details to devices
- Add support for bulk operations

Screenshot 2024-03-12 at 2 38 21 PM


app/controllers/internal_api/v1/users/devices_controller.rb
```diff
@@ -39,7 +32,11 @@ def device
 
     def device_params
       params.require(:device).permit(
-        :device_type, :name, :serial_number, specifications: {}
+        add_devices: [:device_type, :name, :serial_number, :is_insured, :insurance_bought_date, :insurance_expiry_date,
```
nisusam, line 35:
```RUBY
expected_attributes = [:device_type, :name, :serial_number, :is_insured, :insurance_bought_date, :insurance_expiry_date, specifications: [:processor, :ram, :graphics]]

add_devices: expected_attributes,
update_devices: expected_attributes + [:id]
remove_devices: []
```

app/models/device.rb
```diff
@@ -4,15 +4,18 @@
...
-#  name           :string
-#  serial_number  :string
-#  specifications :jsonb
-#  created_at     :datetime         not null
-#  updated_at     :datetime         not null
-#  company_id     :bigint           not null
-#  user_id        :bigint           not null
+#  id                    :bigint           not null, primary key
+#  device_type           :string           default("laptop")
+#  insurance_bought_date :date
```
nisusam, line 9: Is `insurance_bought_date == insurance_start_date`?

nisusam (changes requested):

app/models/device.rb
apoorv1316: using insurance_activation_date now

app/controllers/internal_api/v1/users/devices_controller.rb
apoorv1316: done

nisusam (approved):
Are we maintaining histories of insurance? Anyways, We can tackle it in a new PR.

Rest LGTM!

PR #1703, opened 2024-03-13
Leave balance issue days per month

### LINKED ISSUE #1693: The current leave balance should be calculated as per the leave setting defined by the admin or owner user.
### Description

Currently the leave setting are not being considered for the calculation of leave balance at the start of the year. 

### What is the expected behaviour?

The leave balance should be calculated as per the leave settings that is defined. For E.g. 
2 days per month then until  March the leaves accrued should be 2*3  i.e. 6 days. 

While calculating the total leave balance we need to check the carry forward from the previous year that is also defined in the leave settings for the leave type defined. 

Other changes 
1. Display the number of days first followed by the hours on the leave tiles on the leave manage page. 
2. Change the title  Leave balance until 31st Dec 2024 to show the leave balance until the current month . 

### What is the current behaviour?

Currently, even though we gave defined the leave settings as 2 days per month , the system is calculating the leave for the entire i.e. 2*12  = 24 days. 

### How to reproduce the issue?

1. Login as admin or owner user 
4. Navigate to settings/leave settings
5. Add the leave type PTO and define the number of days/month
6. Click on save 
7. Navigate to the leave management page 
8. View the leave balance

SiddharthShringi:
Fixes #1693 
- Fixed the issue with days per month case.
- Now we are calculating the leave balance based on the user joining date as the start time and the present time as the end time for the current year.
- If the user joining date is the previous year, then we are showing the leave balance based on year start time till present time

https://www.loom.com/share/7bcc76797b7740eea5c674cf81f72ffe?sid=5e4cd9bf-c2c0-4fa9-9363-2283b2da59da

### DIFF
--- a/app/javascript/src/components/LeaveManagement/index.tsx
+++ b/app/javascript/src/components/LeaveManagement/index.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from "react";
 
-import { getYear } from "date-fns";
+import { getYear, format } from "date-fns";
 
 import timeoffEntryApi from "apis/timeoffEntry";
 import Loader from "common/Loader/index";
@@ -95,6 +95,16 @@ const LeaveManagement = () => {
       return <Loader />;
     }
 
+    const getLeaveBlanaceDateText = () => {
+      const presentYear = getYear(new Date());
+      let date = new Date(currentYear, 11, 31);
+      if (currentYear == presentYear) {
+        date = new Date();
+      }
+
+      return `Leave Balance Untill ${format(date, "do 'of' MMM yyyy")}`;
+    };
+
     return (
       <div className="h-full w-full py-6">
         <Header
@@ -106,7 +116,7 @@ const LeaveManagement = () => {
           setSelectedEmployeeId={setSelectedEmployeeId}
         />
         <Container
-          currentYear={currentYear}
+          getLeaveBlanaceDateText={getLeaveBlanaceDateText}
           leaveBalance={leaveBalance}
           selectedLeaveType={selectedLeaveType}
           setSelectedLeaveType={setSelectedLeaveType}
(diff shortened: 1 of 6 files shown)


apoorv1316:
@SiddharthShringi Can you please add loom video for this?

prasanthchaduvula (commented):
Please check the formatting once as mentioned in the comment

apoorv1316 (approved):

SiddharthShringi:
@prasanthchaduvula, yes the organization date format would not affect the leave balance date format.

app/javascript/src/components/LeaveManagement/index.tsx
```diff
@@ -95,6 +95,16 @@ const LeaveManagement = () => {
       return <Loader />;
     }
 
+    const getLeaveBlanaceDateText = () => {
```
nisusam, line 98: `Balance`

app/javascript/src/components/LeaveManagement/index.tsx
```diff
@@ -95,6 +95,16 @@ const LeaveManagement = () => {
...
+      const presentYear = getYear(new Date());
+      let date = new Date(currentYear, 11, 31);
+      if (currentYear == presentYear) {
+        date = new Date();
```
nisusam, line 102: In which case, the current year not match with present year?

spec/services/timeoff_entries/index_service_spec.rb.rb
```diff
@@ -52,7 +53,7 @@
         year: Date.today.year
       }
 
-      service = TimeoffEntries::IndexService.new(params, company, user)
+      service = service = TimeoffEntries::IndexService.new(user, company, params[:user_id], params[:year])
```
nisusam, line 56: Redundant.

Also rename filename to single `.rb`

nisusam (changes requested):

app/javascript/src/components/LeaveManagement/index.tsx
SiddharthShringi: The `currentYear` variable is responsible for current year shown on the leave management page, it will change when we go to previous year leave balance.
And the presentYear is simply present year.

nisusam (approved):

Issue #1695, opened 2024-03-08
Change the button name on Members for project page to "Save changes"


vipulnsward:
This is confusing, as the same button is also used to save changes for deleting members
Change the button name on Members for project page to "Save changes"

![image](https://github.com/saeloun/miru-web/assets/567626/8dd0879d-95a9-4300-a53f-6e2353dc62e1)

PR #1713, opened 2024-03-15
Add members button text replaced

Shruti-Apte:
Closes Issue #1695


nisusam (approved):
Changes LGTM!

I don't know if we want to update this in the system specs.

PR #1727, opened 2024-03-19
Optional holiday lapses

prasanthchaduvula:
### What
- We can't apply for optional holiday more than the defined optional holidays. It will throw the following error message "You have exceeded the maximum number of permitted optional holidays". For example, if the admin defined one optional holiday for one quarter then the employee can apply for more than one optional holiday in the same quarter.

-  We can't apply for two holidays on the same day. It will throw the following error message "You are adding two holidays on the same day, please recheck".


app/models/timeoff_entry.rb
```diff
@@ -57,4 +59,54 @@ def either_leave_type_or_holiday_info_present
...
+          user:).count
+      when :per_quarter
+        start_of_quarter = leave_date.beginning_of_quarter
+        end_of_quarter = leave_date.end_of_quarter
+
+        total_optional_entries = optional_timeoff_entries.where(
+          leave_date: start_of_quarter..end_of_quarter,
+          user:).count
+      when :per_year
+        total_optional_entries = optional_timeoff_entries.where(user:).count
```
nisusam, line 95: `leave_date` should be filtered based on year

nisusam (changes requested):

app/models/timeoff_entry.rb
prasanthchaduvula: We are filtering optional_timeoff_entries from holidays and holidays are unique for each year. The entries are getting filtered based on the holiday year.

app/models/timeoff_entry.rb
```diff
@@ -57,4 +59,54 @@ def either_leave_type_or_holiday_info_present
         errors.add(:base, "Choose either leave type or holiday info, not both")
       end
     end
+
+    def optional_holiday_timeoff_entry
+      return unless holiday_info&.category == "optional" &&
+                  holiday_info&.holiday&.no_of_allowed_optional_holidays.present?
+
+      no_of_allowed_optional_holidays = holiday_info&.holiday.no_of_allowed_optional_holidays
```
nisusam, line 67: DRY `holiday_info&.holiday`

apoorv1316 (approved):

spec/models/timeoff_entry_spec.rb
```diff
@@ -13,6 +19,54 @@
     end
 
     it { is_expected.to validate_presence_of(:leave_date) }
+
+    it "is not valid if it exceeds the maximum number of permitted optional holidays" do # rubocop:disable RSpec/ExampleLength
+      existing_timeoff_entry = create(
```
nisusam, line 24: Use `before` block

nisusam (changes requested):

nisusam (approved):

PR #1762, opened 2024-03-26
Fix specs in timeoff services

apoorv1316:
Specs failing due to hardcoded expectations


prasanthchaduvula:
LGTM Specs got resolved #1757

PR #1750, opened 2024-03-22
Refresh page after adding/editing team member

Shruti-Apte:
Closes #1741 

### What 
- Page refresh after add/edit/delete
- Compensation tab hidden from profile and team details
- Employment and devices tab added on team details

PR #1772, opened 2024-03-28
create invitation presenter

apoorv1316:
Why
- To send the created invitation in JSON response while adding team members.
- To resolve https://github.com/saeloun/miru-web/pull/1750#discussion_r1539194438


app/presenters/invitation_presenter.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def process
+    {
+      id: invitation.id,
+      name: invitation.full_name,
+      firstName: invitation.first_name,
+      lastName: invitation.last_name,
+      email: invitation.recipient_email,
+      role: invitation.role,
+      status: "INVITATION PENDING",
```
prasanthchaduvula, line 18: Instead of status value sending from the backend, let frontend use the whatever message they want to display. Its better to send the key   `accepted_at `

prasanthchaduvula (changes requested):

app/presenters/invitation_presenter.rb
apoorv1316: Shruti asked me to send the same keys that are use in GET requests for teams. LMK if we have to change

prasanthchaduvula: For status, we have to use `accepted_at `. On the frontend, they can display any message as the messages keep on changing.

app/presenters/team_presenter.rb
```diff
@@ -46,29 +46,28 @@ def formatted_invitations
...
-      I18n.t("team.reconfirmation")
+      user_employed_at_current_company && member.confirmed_at?
     end
 
-    def invited_user_status
-      return unless is_current_user_employee
-
-      I18n.t("team.invitation")
+    def invited_user_status(member)
+      user_employed_at_current_company && member.accepted_at
```
prasanthchaduvula, line 61: Use `accepted_at?` return boolean

app/views/internal_api/v1/invitations/_invitation.json.jbuilder
```diff
@@ -0,0 +1,16 @@
...
+json.deep_format_keys!
+
+json.invitation do
+  json.id invitation.id
+  json.name invitation.full_name
+  json.first_name invitation.first_name
+  json.last_name invitation.last_name
+  json.email invitation.recipient_email
+  json.role invitation.role
+  json.status invitation.accepted_at
```
prasanthchaduvula, line 13: Use `accepted_at? ` returns boolean

prasanthchaduvula (changes requested):

prasanthchaduvula (approved):

Issue #1863, opened 2024-06-17
Disable emails by default


sonamvg:
- Disable sending emails for timesheet reminders for all users.

PR #1867, opened 2024-06-19
Add `NotificationPreference` for weekly reminder

nisusam:
- Fix https://github.com/saeloun/miru-web/issues/1863
- By default, preference is set to `false`
- After this PR got merged, we need to run a one-off script to generate `NotificationPreference` records


app/models/company.rb
```diff
@@ -45,6 +45,7 @@ class Company < ApplicationRecord
   has_many :holidays, dependent: :destroy
   has_many :holiday_infos, through: :holidays, dependent: :destroy
   has_many :carryovers
+  has_many :notification_preferences
```
gowsik-ragunath, line 48:
```suggestion
  has_many :notification_preferences, dependent: :destroy
```

app/models/user.rb
```diff
@@ -75,6 +75,7 @@ def initialize(msg = "Spam User Login")
   has_many :custom_leave_users
   has_many :custom_leaves, through: :custom_leave_users, source: :custom_leave
   has_many :carryovers
+  has_many :notification_preferences
```
gowsik-ragunath, line 78:
```suggestion
  has_many :notification_preferences, dependent: :destroy
```

app/services/weekly_reminder_for_missed_entries_service.rb
```diff
@@ -8,7 +8,11 @@ def process
 
     Company.find_each do |company|
       company.users.kept.find_each do |user|
-        check_entries_and_send_mail(user, company)
+        notification_preference = NotificationPreference.find_by(user_id: user.id, company_id: company.id)
+
+        if notification_preference.present? && notification_preference.notification_enabled
+          check_entries_and_send_mail(user, company)
+        end
       end
```
gowsik-ragunath, line 16: Suggestion: I think it would be better if we have this check in SendWeeklyReminderToUserMailer, so that even when we trigger mailer manually, it won't send any email if the user's notification preference is not satisfied.

In future, we can even have this as a before action check in the application mailer, If we plan to use this notification preference for all the mail notifications.

gowsik-ragunath (approved):
LGTM, left few comments and suggestions.

app/models/notification_preference.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+#  fk_rails_...  (company_id => companies.id)
+#  fk_rails_...  (user_id => users.id)
+#
+class NotificationPreference < ApplicationRecord
+  belongs_to :user
+  belongs_to :company
+
+  validates :user_id, presence: true
+  validates :company_id, presence: true
```
apoorv1316, line 30: I think `belongs_to` will automatically validate the presence and it is redundant.

apoorv1316 (approved):

app/models/notification_preference.rb
nisusam: This was added for more readability

PR #1871, opened 2024-07-02
Sidekiq to solid queue migration

apoorv1316:
- Migrated from sidekiq to solid queue for jobs.


Gemfile
```diff
@@ -114,11 +114,8 @@ gem "active_interaction"
 # For stripe payments
 gem "stripe"
 
-# Background job processing adapter
-gem "sidekiq", "~> 7.2"
-
-#  job scheduler extension for Sidekiq
-gem "sidekiq-scheduler"
+gem "mission_control-jobs"
+gem "solid_queue", "~> 0.3"
```
akhilgkrishnan, line 118: Add proper description for these two gems

Procfile
```diff
@@ -1,2 +1 @@
 web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-production}
-worker: bundle exec sidekiq -e production -C config/sidekiq.yml
```
akhilgkrishnan, line 2: don't we wanna run the solidqueue worker on production?

app/controllers/mission_control_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class MissionControlController < ApplicationController
```
akhilgkrishnan, line 3: can you also add a test for this controller?

Gemfile
```diff
@@ -209,7 +206,6 @@ group :test, :ci do
   # Strategies for cleaning databases in Ruby.
   gem "database_cleaner", "~> 2.0"
   gem "hash_dot"
-  gem "rspec-sidekiq", git: "https://github.com/wspurgin/rspec-sidekiq", branch: "main"
```
akhilgkrishnan, line 212: Did you noticed any issue on running the jobs on test?

akhilgkrishnan (commented):
Great work @apoorv1316 🎉,
I've added some minor comments, lemme know your thoughts on that.

Procfile
apoorv1316: I've added it in fly.production.toml . Do we still need to add it here?

Gemfile
apoorv1316: 2 tests were failing after removing this job because they were using the matchers from this job. Now I am using matcher from `rspec-rails` for those tests.

Procfile
akhilgkrishnan: Make it consistent

.env.example
```diff
@@ -27,13 +27,6 @@ GOOGLE_OAUTH_CLIENT_SECRET='google-oauth-client-secret'
 # Application Monitoring
 NEW_RELIC_LICENSE_KEY='replace with newrelic licence key'
 
-# Redis url
```
nisusam, line 30: Keep this until we are getting rid of `Redis` completely

akhilgkrishnan (approved):

PR #1879, opened 2024-07-17
Remove redis

apoorv1316:


scripts/benchmark.rb
```diff
@@ -0,0 +1,22 @@
...
+
+Benchmark.ips do |x|
+  x.report("BulkInvoiceDownloadJob with Redis") do
+    BulkInvoiceDownloadJob.perform_now(
+      invoice_ids,
+      company_logo,
+      download_id,
+      root_url,
+      current_url_options
+    )
```
akhilgkrishnan, line 20: Instead of calling the Job, can you copy all the contents from the job and do a benchmarking.

apoorv1316: I did that and the results are same. Very minor difference. Should I go ahead and delete this file and uninstall redis?

vipulnsward:
Instead of this, could we exploring removing BulkInvoiceDownloadChannel instead if possible?

We could remove ACa completely. 

If not, then we should explore anycable or using hotwire if possible, or just simple polling.

vipulnsward:
@apoorv1316 what's the status on this? Can we wrap this up before picking up other issues?

apoorv1316:
Closing this in favor of #1892

PR #1885, opened 2024-07-29
Sort users by name on time tracking page dropdown

### LINKED ISSUE #1884: Sort users by name on time tracking page dropdown

apoorv1316:
Fixes #1884


keshavbiswa (commented):
It seems like we're duplicating the code, is it possible to extract this out to a reusable module?

apoorv1316:
@keshavbiswa Please review

app/services/employee_fetcher.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module EmployeeFetcher
```
akhilgkrishnan, line 3: Seems like this one is a helper or concern, but its not a service.

apoorv1316: Yeah we can put it in app/services/concerns like we do for model and controllers. But then we have to autoload that. Should I do that?

apoorv1316: moved it to concerns

app/services/concerns/employee_fetching_concern.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module EmployeeFetchingConcern
+  extend ActiveSupport::Concern
+
+  included do
+    def set_employees
+      @employees = admin? ? current_company_users : [current_user]
```
keshavbiswa, line 8: There is an inconsistency of data being sent here. 

If `admin` is true:, we're sending only :id, :first_name, :last_name, but if it's false then we're sending `current_user` to the array. 

Can you check if it's possible to just send id, first_name, last_name or just send the whole object?

config/application.rb
```diff
@@ -17,6 +17,7 @@ class Application < Rails::Application
     # not contain `.rb` files, or that should not be reloaded or eager loaded.
     # Common ones are `templates`, `generators`, or `middleware`, for example.
     config.autoload_lib(ignore: %w(assets tasks))
+    config.autoload_paths += %W(#{config.root}/app/services/concerns)
```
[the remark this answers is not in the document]
apoorv1316: %w does not allow string interpolation. So %W is correctly used here

app/services/concerns/employee_fetching_concern.rb
apoorv1316: done

keshavbiswa (approved):
Looks good, can we replace the slices of data with the full data itself? Seems easier that way and makes it reusable.

PR #1932, opened 2024-11-07
Updated grover config

gowsik-ragunath:
- Added swap to worker machine this will fix the memory issue.

- Grover config changes:
  - `prefer_css_page_size` set to `false` as we don't have any page css property.
  - Set `timeout` to 30 seconds. This will avoid Grover to waiting indefinitively which will also increase the memory usage and cause the server to restart.
  - Added [`--disable-gpu`](https://peter.sh/experiments/chromium-command-line-switches/#disable-gpu) as launch_args to disable the GPU acceleration.
  - set `javascript_enabled: false` as the PDF pages are static.

- Removed `networkidle0` and `networkidle2` as `load` should be enough to get the images from s3 and it will reduce the 500 millisecond wait.
- Fixed the address issue in organization edit form.

### DIFF
--- a/app/javascript/src/components/Profile/Organization/Edit/index.tsx
+++ b/app/javascript/src/components/Profile/Organization/Edit/index.tsx
@@ -133,12 +133,13 @@ const OrgEdit = () => {
       });
       setIsDetailUpdated(true);
     }
-  }, [file]);
+  }, [file, orgDetails]);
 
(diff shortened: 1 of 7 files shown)


keshavbiswa (commented):
LGTM! Haven't tested it yet. Can @apoorv1316 test this on staging and confirm it's working as expected?

PR #1935, opened 2024-11-18
Improve email score

### LINKED ISSUE #1925: Check and improve email delivery score

Shruti-Apte:
Closes #1925 
### What:
- Removed logo attachments
- Some UI fixes (added table display)

### Why:
- logos were being attached as a standard attachment instead of inline 
- flex is not supported in email template


gowsik-ragunath (approved):
LGTM

PR #1942, opened 2024-12-11
Zerobounce integration

apoorv1316:
What
- Added ZeroBounce integration

Why
- To validate user emails before the actual email is sent on sign up.

TODOS
- Add specs.
- Check other places in app where we need to use it.


amolmjoshi93:
Closing this PR as code is merged from https://github.com/saeloun/miru-web/pull/2019 to develop

PR #1956, opened 2025-03-05
Get rid of constants

nisusam:


apoorv1316 (approved):
@nisusam Please check if we can add spec for it. Also verify it on staging. Thanks

PR #1963, opened 2025-04-03
Fixes: Invoice Details for Mobileview

JijoBose:
**What**
- Added Client currency to Mobile view
- Fixed the client currency to show in the client detail page
- Custom validation for base_currency was added when the client and company base currencies were identical.


Shruti-Apte:
Deployed and tested this branch on staging. 
LGTM merging the PR.

Issue #1964, opened 2025-04-03
Invoice number is not available to use after an invoice is deleted


sonamvg:
### Description
The invoice number is not available for use after an invoice is deleted

### How to reproduce the issue?

1. Create an invoice and delete it.
2. Retry to create another invoice for the same client
3. Change the invoice number to the previous number

### What is the current behaviour?

A new invoice with next number is created.
On manually changing the number, he invoice number is shown as used and is not available for a new invoice

### What is the expected behaviour?

After an invoice is deleted, the invoice number should be available to use.

PR #1965, opened 2025-04-04
The invoice number is not available for use after an invoice is delete bug fix

minakshin:
Github task - https://github.com/saeloun/miru-web/issues/1964

Changes

1. Invoice after_discard updating invoice_number with prefix
 2. Added active scope for Invoice for new invoice generation

https://github.com/user-attachments/assets/ee6ae5d5-1406-4578-b139-db29c66a9e6d

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -351,6 +351,7 @@ GEM
     memoist (0.16.2)
     mini_magick (4.12.0)
     mini_mime (1.1.5)
+    mini_portile2 (2.8.8)
     minitest (5.25.4)
     mission_control-jobs (0.2.1)
       importmap-rails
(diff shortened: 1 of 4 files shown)


apoorv1316 (commented):
@minakshin I would also write a spec for verifying that I am able to use the invoice number that was used in the discarded invoice.

apoorv1316:
@Shruti-Apte Please get it deployed on staging and test it.

minakshin:
> @minakshin I would also write a spec for verifying that I am able to use the invoice number that was used in the discarded invoice.

Updated test cases for spec to reuse discarded invoice number

apoorv1316 (approved):

PR #2006, opened 2025-09-19
Bump version to 1.9.1

shalapatil:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@saeloun/miru-web",
-  "version": "1.9.0",
+  "version": "1.9.1",
   "dependencies": {
```
apoorv1316, line 4: this points to the next upcoming version so it would se something like 1.9.2.

apoorv1316 (approved):

PR #2384, opened 2026-07-09
Security, currency correctness, dark-mode & feature fixes from a deep audit

vipulnsward:
## Summary

A deep audit of the app (security, money/data-integrity, frontend correctness, design, and features) turned into a set of focused, individually-reviewed fixes. Everything is on one branch as ten self-contained commits, each with regression tests and, for security/money changes, an adversarial review pass.

## What's in it

**Security (`0cd842521`)**
- **CRITICAL** cross-tenant IDOR: bulk invoice download loaded `Invoice.where(id:)` with no company scope — any signed-in user could download other companies' invoice PDFs. Now scoped end to end (+ migration adds `company_id` to the status table).
- Client-portal users could list/download *every* invoice in the company; now scoped to invoices of a client they belong to (the general index no longer grants client access — they use the dedicated client controller).
- Unauthenticated payment-success endpoint leaked invoice financials by sequential id; now keyed on the opaque `external_view_key`.
- Employees could log time-off against any coworker; now self-restricted.
- Web OTP/TOTP verify endpoints were missing the brute-force throttle their mobile twins already had.
- `backup.yml` ran unpinned `@master`/`@latest` actions while holding prod DB + R2 creds — SHA-pinned + `permissions: contents: read`.
- bundler-audit: removed two dead ignores (one risked hiding a *critical* advisory) and corrected the webrick rationale.

**Money & reliability (`950bef6fc`)**
- Back-dated invoices/payments/expenses were stamped with *today's* FX rate; now resolve the rate for the requested date.
- FX HTTP calls had no timeout (could hang a save on the request thread) — added 5s timeouts.
- Silent 1:1 FX fallback on payments now reports to Sentry with context.
- Payments report and expense summaries now total in base currency instead of summing mixed currencies / dropping foreign-currency rows (+ `base_currency_amount` migration on expenses + backfill task).
- Payments are now soft-deletable and an invoice with recorded payments can't be hard-destroyed (bulk delete reports the ones it skipped).
- `Settle` is transactional; Stripe checkout fulfillment locks the invoice to close the double-settle race.
- Timesheet week windows and the reported timezone use the company timezone.

**Frontend correctness (`57d8afe04`)**
- Expense date saved off-by-one for non-UTC users (raw dayjs object → `YYYY-MM-DD`).
- Guarded a `JSON.parse` that crashed the Invoices list on corrupt localStorage; guarded the aging-report totals against missing data; surfaced previously-swallowed timesheet/clients errors; made the receipt-upload target keyboard-operable.

**Design (`6d6405c88`, `f96ad24dd`, `7beb60077`, `686916091`)**
- Brand: `--primary` was near-black; set it (and `--ring`, sidebar, charts) to the DESIGN.md purple `#5B34EA`, removed the rejected teal, and swapped the 72 hardcoded `#5E58F1` usages for the token. Browser-verified.
- Dark mode: ~270 hardcoded gray/white utilities → semantic tokens so they invert; error text → `text-destructive`. Browser-verified both themes.

**Features (`b889a1cd2`)**
- Wired the dead workspace switcher to the existing endpoint; implemented the payments-report PDF (was shipping a file containing the literal string "PDF generation not yet implemented"); added an expense audit trail.

**Sentry fixes (`067c2d98a`)**
- Malformed multipart requests now return `400`, and scanner POSTs no longer fall through to the SPA.
- Logout ignores malformed bearer tokens while still revoking valid JWTs.
- Client uniqueness races return `422`, and the clients index uses preloaded invoices instead of one latest-invoice query per client.
- Missing workspaces return `404`, and absent optional-holiday allowances count as zero.

## Testing & review
- Prior audit batch: full backend suite green (2973 examples, 0 failures). Sentry batch: 85 focused examples pass, RuboCop passes on all 17 changed files, and affected browser routes render without React console crashes. The wider suite still has 12 unrelated currency/system failures reproduced on the unchanged PR head.
- Design and the PDF/Ferrum output verified in a real browser (Chromium + WebKit) / via runner.
- Three adversarial red-team passes (security, money/correctness, frontend/design). They confirmed the headline fixes hold under attempted bypass and surfaced a handful of issues, all addressed in `a48fe208b`: a residual intra-company IDOR in bulk previous-employments (an employee could rewrite a coworker's record), a missing FX recompute-guard on Payment, payments listings not scoped to `.kept`, a switch-thumb contrast regression from the token sweep, and stale UI after a partial bulk-delete.

## Known follow-ups (intentionally not in this PR)
- **Stripe idempotency**: fulfillment still dedupes on `invoice.paid?`. This PR closes the concurrency race (`with_lock`), but a full fix should dedupe on the Stripe event/payment-intent id (+ a unique index) so a payment isn't dropped when the invoice is already paid, and partial-payment retries can't double-record.
- **FX**: for back-dated rows only the dated frankfurter provider runs (the latest-only providers are correctly skipped), so pairs it doesn't cover fall to the 1:1-with-Sentry-warning fallback (the behaviour chosen for this PR). Run `exchange_rates:recalculate_expenses` / `:recalculate_payments` after deploy to backfill `base_currency_amount` on existing rows.
- **Auth hardening**: add a per-token attempt lockout on the internal OTP path and a signup/reset-password throttle; add a unique index on `bulk_invoice_download_statuses.download_id`.
- **`darkMode: "class"`** is missing from `tailwind.config.js`, so the app's ~60 existing `dark:` Tailwind utilities follow the OS, not the in-app toggle. One-line fix, but warrants an app-wide dark-mode QA pass; the status-badge dark variants are deferred until it lands. A couple of stray hardcoded colours (`#F5F7F9`, `bg-purple-50/50`) also remain unswept.
- List pagination (payments/clients) needs coordinated frontend work; Xero is greenfield (its own PR).

## Summary by CodeRabbit

* **New Features**
  * Added automated base-currency conversion for expenses and payments, with base-currency totals reflected in payment reports and PDFs.
  * Added scheduled daily exchange-rate updates and expanded bulk invoice-download scoping.

* **Bug Fixes**
  * Improved bulk invoice deletion to block undeletable invoices and return clear errors.
  * Tightened access controls for invoices and time-off/employment actions.
  * Enhanced sign-out token revocation and handling of malformed requests.

* **Style**
  * Updated the app’s theme styling (card backgrounds, border/divider colors, and primary color usage) across tables, forms, reports, and print views.


moorebernal-blip (approved):

moorebernal-blip (commented):
Update this now

PR #2396, opened 2026-07-23
fix: allow editing and deleting time off entries

### LINKED ISSUE #2393: Not able to edit or delete time off entries
### What is the expected behaviour?
We should be able to delete or edit time off entries same as to normal timesheet entry.

### What is the current behaviour?
Not able to edit or delete time off entries

### How to reproduce the issue?

1. create time off entry
2. try to edit or delete it
3.

vipulnsward:
Fixes #2393.

## Root cause

`CompanyDateFormattingService` always serializes `leave_date` with **dot** separators (`%m.%d.%Y` / `%d.%m.%Y` / `%Y.%m.%d`), while every frontend strict `dayjs` parse only listed dash/slash formats. Editing a time off entry therefore:

- poisoned `selectedFullDate` with the raw dotted string (`handleEditTimeoff` fallback),
- sent `leave_date: "07.23.2026"` to the API → **422 "Leave date can't be blank"**,
- left the stale leave card in the list and injected the raw update response as a phantom *timesheet* card whose delete button hit the wrong API.

Reproduced end-to-end in the browser before fixing (8h PTO edited to 4h showed "12h, 2 entries").

## Fix

- New shared `buildDateParseFormats(companyDateFormat)` helper — company format and its dot variant first, then canonical fallbacks — used by TimeTracking, TimeEntriesDisplay, and TimeoffForm (previously three drifting inline lists; the duplicate-entry payload path also used a non-strict parse that produced literal `"Invalid Date"`).
- Timeoff update flow now always refetches (mirrors create) instead of filter+relocate of raw response JSON; form closes unconditionally on 2xx; removed a double `fetchEntriesOfMonths` fetch that could clobber the visible week.
- `custom_leave_id` exposed in the time-tracking / timesheet leave payloads so custom-leave entries are editable too.

## Security hardening (from adversarial review)

- **IDOR closed**: `OverviewService` now clamps `user_id` at entry — non-admins always get their own entries. Previously any employee could read a co-worker's leave dates, durations, and notes via `GET /api/v1/timeoff_entries?user_id=<victim>`.
- **Cross-tenant oracle closed**: admin lookups are company-scoped (`current_company.users.find`), so foreign-company ids 404 instead of returning fabricated balances; `BalanceSummaryService#user_joined_date` no longer raises on a blank `user_id` (`User.find(nil)` 404'd the whole index for admins, breaking Leave Management's first load).

## Verification

- 61 specs green across `spec/requests/api/v1/timeoff_entries`, `spec/services/timeoff_entries`, `spec/requests/api/v1/timesheet_entry`, including new regression specs: admin without `user_id` (200 + leave balance), employee passing another user's id (only own entries returned), admin passing foreign-company id (404).
- Verified in the browser with Playwright: create → edit (PUT sends ISO `leave_date`, single card updates in place, form closes) → delete (card removed, empty state) with zero console errors.
- Reviewed by testing/maintainability/performance specialist passes + adversarial review; all fixable findings addressed (two pre-existing items deferred: UserContext date-format key normalization, `company.date_format` whitelist validation).
- Vite build, ESLint, and RuboCop clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for company-specific date formats across time tracking and time-off workflows.
  * Time-off and timesheet API responses now include custom leave identifiers.

* **Bug Fixes**
  * Improved date handling when creating or editing time-off entries.
  * Edited time-off entries now refresh reliably and update displayed dates correctly.
  * Improved access scoping so time-off data is limited to the appropriate company and employee.

### DIFF
--- a/app/javascript/src/components/TimeTracking/TimeEntriesDisplay.tsx
+++ b/app/javascript/src/components/TimeTracking/TimeEntriesDisplay.tsx
@@ -6,7 +6,7 @@ import customParseFormat from "dayjs/plugin/customParseFormat";
 import { Button } from "../ui/button";
 import { Card, CardContent } from "../ui/card";
 import { Badge } from "../ui/badge";
-import { minToHHMM } from "../../helpers";
+import { buildDateParseFormats, minToHHMM } from "../../helpers";
 import { i18n } from "../../i18n";
 import { useUserContext } from "../../context/UserContext";
 import { Roles } from "../../constants";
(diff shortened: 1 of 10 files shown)


moorebernal-blip (commented):
Mk

